PR #252, opened 2021-06-30
Welcome modal loading update

### LINKED ISSUE #204: Update content of the modal to explain what is being loaded while service worker loads
**Describe the solution you'd like (import visuals when possible)**

- [ ] Update content of the modal to explain what is being loaded and why it may take some time

Mockup [here](https://www.figma.com/file/4J9xyfF5v4B98MuKVO0Y8V/Labelflow-mockups?node-id=2256%3A12469)

crubier:
Welcome modal loading update

 - Move import tutorial in the welcome modal (With fixed image ids)
 - Fixed some bad race conditions on dexie DB

Closes #204


crubier:
Wrong Browser

Screenshot 2021-08-30 at 14 13 04

Wrong browser AND error when initializing

Screenshot 2021-08-30 at 14 13 00

Loading nominal 

Screenshot 2021-08-30 at 14 13 09

Welcome nominal

nicOwlas:
- Brave: ❌ The modals are not working. Loading the application for a very long time, then having the error screen, then it's blinking ([see screen recording](https://user-images.githubusercontent.com/16015322/131462668-6c7158e1-1166-461c-8a24-518317569e76.mov))
- Firefox ✅ App loads very fast (less than a sec). Great UX.
- Chrome: ✅ App loads very slowly (8 sec). It feels long.
- Safari: ✅ App loads very slowly (9 sec). It feels long.

crubier:
The blinking error loop was due to a typo, I missed a `/` character https://github.com/Labelflow/labelflow/pull/252/commits/bb99d76d296ddd10f8c498bc1a99e2bf1fd7afbf

crubier:
But I can't replicate, it works for me on brave with shield, on localhost...

RaynalHugo:
This story seems to be broken
https://www.chromatic.com/component?appId=60a646559841590039471afc&name=web%2Fapp+lifecycle%2Fwelcome+modal&buildNumber=1513&specName=Default

typescript/web/src/components/app-lifecycle-manager/welcome-modal/welcome-modal.tsx
```diff
@@ -1,185 +1,357 @@
...
+          refetchQueries: [
+            "getDatasets",
+            "getDatasetById",
+            "getAllImagesOfADataset",
+            "getDataset",
+            "countLabelsOfDataset",
+            "getLabelClassesOfDataset",
+            "image",
+            "getImageLabels",
+          ],
```
RaynalHugo, line 133: This looks error-prone. I am afraid we will forget to add a query here one day, or that we will rename one without seeing that the name was important.

Can't we perform a cache update rather than this?

typescript/web/src/components/app-lifecycle-manager/welcome-modal/welcome-modal.tsx
```diff
@@ -1,185 +1,357 @@
...
+        (name === "chrome" && versionNumber >= 45) ||
+        (name === "safari" && versionNumber >= 14)
+      ) ||
+      os === "BlackBerry OS" ||
+      os === "Windows Mobile" ||
+      os === "Windows 3.11" ||
+      os === "Windows 95" ||
+      os === "Windows 98" ||
+      os === "Windows 2000" ||
+      os === "Windows ME"
```
RaynalHugo, line 188: Small proposal, feel free to ignore

You could have done something like: 
```typescript
const unsupportedOSs = ["BlackBerry OS", /* ... */ , "Windows ME"];
const isOSSupported = unsupportedOSs.includes(os);
```

crubier:
I just tried 3 times in a row with brave and it worked every time:


https://user-images.githubusercontent.com/2954572/131475033-b2d3c9de-cee2-4c4a-be0a-db3b3db05548.mov


https://user-images.githubusercontent.com/2954572/131475067-ceb806bb-7146-42a5-9577-3c5144e5911b.mov


https://user-images.githubusercontent.com/2954572/131475078-31c2d9aa-abda-4fa4-abcc-e0ab19246ce7.mov

crubier:
However something that probably slows down a lot the loading is that we preload images (normal), but images on the website are HUGE !! There's like 4Mb of images on the website to load each time, not good. 


Screenshot 2021-08-31 at 10 05 12

typescript/web/src/components/app-lifecycle-manager/welcome-modal/welcome-modal.tsx
```diff
@@ -1,185 +1,357 @@
...
+  initialBrowserWarning?: boolean;
+  initialIsLoadingWorkerAndDemo?: boolean;
+  initialBrowserError?: Error;
+}) => {
+  const router = useRouter();
+
+  const handleError = useErrorHandler();
+  const client = useApolloClient();
+
+  const [{ hasUserTriedApp }, setHasUserTriedApp] = useCookies([
```
RaynalHugo, line 164: Nitpicking, feel free to ignore

I would rename `setHasUserTriedApp` into `setCookie`. 
Later you do `setHasUserTriedApp("hasUserTriedApp", "false");` while `setCookie("hasUserTriedApp", "false");` could be more accurate.

typescript/web/src/components/app-lifecycle-manager/welcome-modal/welcome-modal.tsx
```diff
@@ -1,185 +1,357 @@
...
+      isOpen={
+        shouldShowBrowserErrorModal ||
+        (!shouldShowBrowserErrorModal && shouldShowBrowserWarningModal) ||
+        (!shouldShowBrowserErrorModal &&
+          !shouldShowBrowserWarningModal &&
+          shouldShowLoadingModal) ||
+        (!shouldShowBrowserErrorModal &&
+          !shouldShowBrowserWarningModal &&
+          !shouldShowLoadingModal &&
+          shouldShowWelcomeModal)
```
RaynalHugo, line 315: I don't know if you made this on purpose to be explicit and some conditions are useless, or if the case where `shouldShowBrowserErrorModal` and `shouldShowBrowserWarningModal` are both true and we specifically want to close the modal here.

If that's the second option, you can ignore this message. 
Else: 

I guess you added this to do a kind of sanity check on whether or not we are in an incorrect state.
For example `!shouldShowBrowserErrorModal && !shouldShowBrowserWarningModal && shouldShowLoadingModal`.

IMO, I think it would be better to split the sanity check from the `isOpen` condition. 
I would do something like 
```typescript
if (shouldShowBrowserErrorModal && shouldShowBrowserWarningModal || allTheOtherIncorrectCombinaisons ) {
  throw new Error("Invalid State, this shouldn't happen");
}
```
And then you can safely do something like 
```typescript
const isOpen = shouldShowBrowserErrorModal | shouldShowBrowserWarningModal | shouldShowLoadingModal | shouldShowWelcomeModal;
```

This way, you could "properly" handle the incorrect case. With the current implementation, you handle this case by not opening the modal.

typescript/web/src/components/app-lifecycle-manager/welcome-modal/welcome-modal.tsx
```diff
@@ -1,185 +1,357 @@
...
+
+        // Nominal welcome modal on first vist after loading
+        if (shouldShowWelcomeModal) {
+          return (
+            <Welcome onClickSkip={handleSkip} onClickNext={handleGetStarted} />
+          );
+        }
+
+        return null;
+      })()}
```
RaynalHugo, line 354: Nitpicking:

This could be a component instead of an IIFE.

typescript/web/src/pages/index.tsx
```diff
@@ -30,18 +34,13 @@ const IndexPage = ({ cookie }: { cookie: string }) => {
...
-  props: { previewArticles: Omit<Article, "content">[] };
-}> => {
-  const previewArticles = (await getAllArticles({ limit: 3 })) || [];
-  const parsedCookie = useCookie(context);
+export const getServerSideProps: GetServerSideProps = async (context) => {
+  const parsedCookie = new Cookies(context.req.headers.cookie);
 
-  if (parsedCookie.get("hasUserTriedApp")) {
+  if (parsedCookie.get("hasUserTriedApp") === "true") {
+    console.log("REDIRECTT");
```
RaynalHugo:
```suggestion
  if (parsedCookie.get("hasUserTriedApp") === "true") {
```

crubier (commented):
Reviewed myself just to check, seems ok

RaynalHugo (approved):
I made a few suggestions but mostly nitpicking. Feel free to ignore them if you don't find them relevant.

Cool work 👏

typescript/web/src/components/app-lifecycle-manager/welcome-modal/welcome-modal.tsx
```diff
@@ -1,185 +1,357 @@
...
+      imagesAggregates {
+        totalCount
+      }
+      labelsAggregates {
+        totalCount
+      }
+      labelClassesAggregates {
+        totalCount
+      }
+    }
```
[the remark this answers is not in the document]
crubier: No

RaynalHugo:
> > This story seems to be broken
> > https://www.chromatic.com/component?appId=60a646559841590039471afc&name=web%2Fapp+lifecycle%2Fwelcome+modal&buildNumber=1513&specName=Default
> 
> It's not broken, it's supposed to be empty ! It's the modal in it's default state, which should be closed ! The story allows to catch bugs where it opens when it should not. I'll rename it.

Ah ok 
Thanks for the rename in d7235e2b5c96aad31313cb45a9aea60ea044ceca. I think that's more explicit.

typescript/web/src/components/app-lifecycle-manager/welcome-modal/welcome-modal.tsx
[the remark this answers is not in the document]
crubier: But it's more as a safety feature. If this query fails then I prefer to catch it in the modal, and tell the guy there's a problem with his browser, instead of letting him go and let him have an error later on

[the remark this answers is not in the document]
RaynalHugo: Ah ok! this makes sense.

typescript/web/src/components/app-lifecycle-manager/welcome-modal/welcome-modal.tsx
crubier: We discussed this with @frinyvonnick , but this is way less error prone (and 10x shorter) than a cache update (sadly). Cache updates are huge and as coupled to other pages as this.

crubier: We can discuss during DSM

typescript/web/src/components/app-lifecycle-manager/welcome-modal/welcome-modal.tsx
crubier: Yes. I could have written it that way, your solution is good. 

But I'll ignore it for now since that would add some indirection. Here the condition is obvious, don't have to scroll on top of the file to have the list of unsupported stuff.

typescript/web/src/components/app-lifecycle-manager/welcome-modal/welcome-modal.tsx
crubier: I know this is weird, but there would be duplication, we have `setHasUserTriedApp` and `setTryDespiteBrowserWarning` which would BOTH be called `setCookie`. And they are not interchangeable, their types are not compatible. So don't really have a choice here.

typescript/web/src/components/app-lifecycle-manager/welcome-modal/welcome-modal.tsx
crubier: If `shouldShowBrowserErrorModal && shouldShowBrowserWarningModal`, we do want to show the modal, and that's what is done here (the first clause of the disjunction is true, so the disjunction is true).

The reason why I added such a complicated condition is that: 

1. There is A LOT of cases (I counted at least 96 different cases in an excel spreadsheet !), (Cookie, error, browser warning, error on the page behind the modal, 404 behind the modal, user clicked some button or not, etc...) so better be explicit and not fuck around.
2. I dont want the modal to be displayed without having a Content
3. So the condition for `isOpen` is exactly (done in a mechanical way) the conditions that display stuff in its content. (the Immediately-invoked Function Expressions inside the modal content)
4. In this Immediately-invoked Function Expressions in the content, we use a function with early returns, so I just combine the conditions of each early return in a disjunction, and use that as `isOpen`

typescript/web/src/components/app-lifecycle-manager/welcome-modal/welcome-modal.tsx
crubier: But then it would separate (in the code) this logic from the logic of the `isOpen` condition. See my previous answer about the condition. Both sides are related, so better to keep them together.

typescript/web/src/pages/index.tsx
crubier: Done

typescript/web/src/components/app-lifecycle-manager/welcome-modal/welcome-modal.tsx
RaynalHugo: You could define the child component right next to where you would define the isOpen const.
I know that this results in exactly the same thing from React's point of view. I think it's just more "react standard".

But that's fine like this if you prefer.

geoffreyvancassel:
I have changed image sizes. Actually the screenshot home-screenshot3 and home-screenshot4 are not used anymore.

nicOwlas:
### Brave
- [I faced the issue again on a first try](https://user-images.githubusercontent.com/16015322/131504756-0653508d-5f61-460e-9e81-aa468c0d6199.mov). A bit different this time. When clicking on see debug info and click previous it works.
- Trying a second time worked without any issue... Loading is 1-2 sec faster. Down to 6sec on Brave.

### Firefox
- It's lightning fast in comparison to Brave ([recording](https://user-images.githubusercontent.com/16015322/131506212-e93055eb-9c3e-4e99-b153-19435c83d556.mov)). The getting ready modal is on screen for less than 0.2-0.3 sec. But maybe too fast (see next point)
- I land on a "page not found" empty state when I click on start tutorial. Like if the image was in fact not ready yet. Talked with Hugo IRL about it.

nicOwlas:
> It's lightning fast in comparison to Brave (recording). The getting ready modal is on screen for less than 0.2-0.3 sec. But maybe too fast (see next point)
> I land on a "page not found" empty state when I click on start tutorial. Like if the image was in fact not ready yet. Talked with Hugo IRL about it.

Fixed with Hugo. Clearing everything made it work. Still lightning fast. The difference with other browsers is impressive. Definitely the one to use for demos.

Issue #317, opened 2021-07-27
Implement CRU of workspaces


nicOwlas:
**Tasks**

- [x] Write type definitions
- [x] Graphql fields:  `name`, `datasets`, `type` (`local` | `online`), `createdAt`, `id`, `updatedAt`
- [x] Update `plan` names (`community`, `starter`, `pro`, `enterprise`)(todo: change plan names in graphql fields)
- [ ]  Create resolvers to read / create / list (online workspaces user is linked to only) / update  (deleted can be done in another issue, we need to think what it implies for the billing if a user deletes a workspace. Server side: CRU resolvers to be implemented. On Service Worker side Read/List returns a single list element. No create or update.
- [x] Add to Dataset table definition the column workspace id
- [x] Make sure to remove the global uniqueness on `slugs` of datasets  (and findUnique etc..). Unicity is on datasets slugs within a workspace.
- [x] Create a join table to link the user table and the workspace table (many to many relationship as per the wireframe of [teams](https://www.figma.com/file/4J9xyfF5v4B98MuKVO0Y8V/Labelflow-mockups?node-id=1850%3A15154) and [workspaces](https://www.figma.com/file/4J9xyfF5v4B98MuKVO0Y8V/Labelflow-mockups?node-id=1865%3A12984) done in the Design Sprint)

Workspace owners.
- [x] ~Add a field `workspaceOwner` and store the userId who created the workspace~
- [x] Add a column `role` on the join table linking users to workspaces. Write `owner` for the user who created the workspace.

blocked by #382

PR #354, opened 2021-08-18
Feature/#317 cru workspaces

GuilhemdV:
WIP for #317 

I only committed the updated the graphql typedefs, the resolvers changes are stashed on my computer because they were not finished


crubier (approved):
Looks good to me

crubier:
Tested live and merged

PR #369, opened 2021-08-20
Bug/#360 cant modify invalid geometry

### LINKED ISSUE #360: Cant modify invalid geometry
**Describe the bug**

Cant modify invalid geometry

Error Updating Label: Each LinearRing of a Polygon must have 4 or more Positions.

**To Reproduce**
1. Create an invalid geometry
2. Try to modify it

Screenshot 2021-08-18 at 22 20 27

**Solution**

- [ ] Correct the function making invalid geometries valid ([check that line](https://github.com/Labelflow/labelflow/blob/55088873d875c156e9ee14f31e38ef72dd77477d/typescript/common-resolvers/src/utils/get-bounded-geometry-from-image.ts#L11))

etiennedupont:
# Feature

## Work performed

- Use `MultiPolygon` instead of `Polygon` when creating polygons
- Handle MultiPolygon in `Coco export`

## Resolved issues

Closes #360

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,128 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+We as members, contributors, and leaders pledge to make participation in our
+community a harassment-free experience for everyone, regardless of age, body
+size, visible or invisible disability, ethnicity, sex characteristics, gender
+identity and expression, level of experience, education, socio-economic status,
+nationality, personal appearance, race, religion, or sexual identity
+and orientation.
+
+We pledge to act and interact in ways that contribute to an open, welcoming,
+diverse, inclusive, and healthy community.
+
+## Our Standards
+
+Examples of behavior that contributes to a positive environment for our
+community include:
+
+* Demonstrating empathy and kindness toward other people
+* Being respectful of differing opinions, viewpoints, and experiences
+* Giving and gracefully accepting constructive feedback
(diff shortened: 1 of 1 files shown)


typescript/common-resolvers/src/format-coco/coco-core/converters.ts
```diff
@@ -67,6 +71,36 @@ const convertLabelClassesToCocoCategories = (labelClasses: DbLabelClass[]) => {
...
+  return geomReduce(
+    flatten(geometry),
+    (segmentations: number[][], currentGeometry) => [
+      ...segmentations,
+      coordReduce(currentGeometry, coordToSegmentationReducer, []),
+    ],
+    []
+  );
+};
+
```
etiennedupont, line 103: Since we know that we're manipulating only MultiPolygons, I was wondering whether we could avoid using `flatten`, `geomReduce`, `coordReduce` by just making a reduce in `geometry.coordinates`, and keeping the first linear ring for each polygon.

crubier: I'd prefer to keep some geometry-agnostic code with `geomReduce` and others, it's a better practice I feel.

etiennedupont: I agree, still, is `flatten` strictly required l95 @ShauniArima ? Since I see `_multiFeatureIndex` l83 which makes me think that a multipolygon would be iterated on each enclosed polygon?

PR #375, opened 2021-08-25
Add versioning and GitHub releases

RaynalHugo:
# Work performed

- Use [release-it](https://github.com/release-it/release-it) to handle our release.
- Automatically run `yarn release minor` on CI when there is a change on `main`

#### 🚨⚠️ Important ⚠️🚨 
Changelog/release messages are base on the PR name when it is merged. Make sure to give meaningful names to your PR before merging them.

Before merging this PR, please ping me or delete the https://github.com/Labelflow/labelflow/releases/tag/0.1.0 and its associated tag. 

## Results

If you run `yarn release [ major | minor | patch ]` at root

1. It updates all packages versions according to the kind of release you chose
2. It generates a changelog using [auto-changelog](https://github.com/cookpete/auto-changelog)
3. It updates the license to last for 2 years after the date of the merge
4. It commits all those changes
5. It creates a tag with the release version
6. It pushes the commit and the tag
7. It creates a GitHub release associated with the git tag (for this to work you need to provide an env var named: GITHUB_TOKEN)

Also, the CI pipeline is now split into 2 workflows : 
- `CI` which only runs on merge requests
- `Release` which is only triggered on `main`. 

Here is a full release pipeline: 
https://github.com/Labelflow/labelflow/actions/runs/1167158342

Here is an example of Github release
https://github.com/Labelflow/labelflow/releases/tag/0.1.0

## Potential enhancements to plan
- Use another changelog generator (like [github-changelog-generator](https://github.com/github-changelog-generator/github-changelog-generator)) to be able to change PR name after merges
- Decide if the release should be `major`, `minor` or `patch` according to a Github label on the PR?


typescript/release/src/update-packages-version.ts
```diff
@@ -0,0 +1,61 @@
...
+            __dirname,
+            `../../${packageName}/package.json`
+          )}`
+        );
+      }
+    })
+  );
+};
+
+main();
```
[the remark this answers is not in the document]
crubier: @RaynalHugo you probably want to use this https://github.com/rwjblue/release-it-yarn-workspaces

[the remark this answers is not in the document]
RaynalHugo: Solved by moving `release-it` command to the web package.

.github/workflows/release.yml
```diff
@@ -0,0 +1,276 @@
...
+      - run: npx vercel logs ${{steps.vercel.outputs.preview-url}} --token ${{ secrets.VERCEL_TOKEN }}
+      - name: update deployment status
+        uses: bobheadxi/deployments@v0.4.3
+        if: always()
+        with:
+          step: finish
+          token: ${{ secrets.GITHUB_TOKEN }}
+          status: ${{ job.status }}
+          deployment_id: ${{ steps.deployment.outputs.deployment_id }}
+          env_url: ${{steps.vercel.outputs.preview-url}}
```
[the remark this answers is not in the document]
RaynalHugo: Done in 5df80f27dbec624008d507422e17b23609a3a4e7

Example of a full release workflow: https://github.com/labelflow/labelflow/actions/runs/1193721407

.github/workflows/ci.yml
```diff
@@ -185,22 +162,29 @@ jobs:
           CI_BRANCH: ${{ github.ref	}}
           CI_BRANCH_BASE: refs/heads/main
 
-  deploy-strapi-main:
-    name: Deploy Strapi on Heroku production
-    if: contains( github.ref, 'main')
+  deployment-chromatic-pr:
```
crubier, line 165: Can you explain the changes here, why chromatic is impacted for example ?

crubier (commented):
I don't understand all the changes in github actions, but otherwise it's good

.github/workflows/ci.yml
crubier: Ok I dived a bit deeper in it. I'll accept it even though I have these questions:

 - Why does deployment-chromatic-main depends on bump-version ? 
 - Why does deploy-strapi depends on bump version ?
 
 I understand why deploy-web-app depends on bump-version of the web-app, but not sure about the others

crubier (approved):

.github/workflows/ci.yml
RaynalHugo: `bump-version` performs changes, commit them and pushes them. 

I made a small change on the `deploy-X-main` jobs. The checkout action now doesn't checkout the commit which triggered the pipeline (default behaviour, but the latest commit from main). This way it checkouts the commit created by `bump-version`. 

I did this to make sure that the deployed App version would correspond to the commit bumping the version. 

Maybe it's overkill for strapi and chromatic. I think it would make it more straightforward for debugging. If someone complains about something being bugged in version X, you just have to find the corresponding commit on Strapi/Chromatic.

RaynalHugo: If you don't think that this is necessary, I can easily change that. I will also speed up the CI.

Issue #388, opened 2021-08-30
Update tutorial images


nicOwlas:
**Tasks**

- [x] Update tutorial images (files attached)
- [x] Set a fix image ID for the tutorial images

[Archive.zip](https://github.com/Labelflow/labelflow/files/7077397/Archive.zip)

RaynalHugo:
Done in #392

PR #392, opened 2021-08-30
Enhancement/update tutorial images

RaynalHugo:
# Feature

## Work performed

- Updated images
- Made image ids given as inputs of the mutation. They can be imported as ` import { tutorialImageIds } from "@labelflow/common-resolvers";`

## Results

tutorial-dataset

## Caveats

- Still some blinks due to the lack of image metadata prefetch
- I didn't really hardcode the images Ids. They are generated at runtime. This wont really make sense on a distant backend but I am not sure we will create a demo dataset the same way.

## Resolved issues

#388


typescript/common-resolvers/src/data/dataset-tutorial.ts
```diff
@@ -6,13 +6,15 @@ const origin =
     ? globalThis.location?.origin
     : "https://labelflow.ai";
 
-export const tutorialImageUrls = [
-  `${origin}/static/img/tutorial-image-1.jpg`,
-  `${origin}/static/img/tutorial-image-2.jpg`,
-  `${origin}/static/img/tutorial-image-3.jpg`,
-  `${origin}/static/img/tutorial-image-4.jpg`,
+export const tutorialImageDetails = [
```
crubier, line 9: Can you remove all the code changes from this PR please ? We said I'd do this in my PR, and I did already in that commit https://github.com/Labelflow/labelflow/pull/252/commits/c8d6f54e4006d3cc8be8cd2a0f2516a9afa86c12

RaynalHugo: It was specified here https://github.com/Labelflow/labelflow/issues/388 to do it ... 
Ok reverting the changes ...

RaynalHugo: I removed the changes. This PR now only contains the images changes.

crubier (approved):

PR #406, opened 2021-09-01
change meta image

geoffreyvancassel:
# Feature


crubier (changes requested):
You can't delete that file. It's used in typescript/web/src/components/meta.tsx 
You need the meta image in 2 different formats

crubier (changes requested):
You removed the seo images !

crubier (approved):

PR #411, opened 2021-09-01
Enhancement/lighthouse fixes

crubier:
# Feature

## Work performed

- Made lighthouse happy
- This includes activating the Service worker on the website already (with a "no Modals" mode)

## Results

 - Better lighthouse score
 - Faster initial load once on the app (it's almost instant now)

## Problems encountered

Not much


crubier:
Before

Screenshot 2021-09-01 at 23 29 44

After

Screenshot 2021-09-02 at 00 44 54

crubier:
After

Screenshot 2021-09-02 at 00 44 54

PR #417, opened 2021-09-02
Make a more generic version of the coco exporter

### LINKED ISSUE #387: Add YOLO export format
**Tasks**

Create a new resolver for YOLO:
- [x] Create one .txt file per image. Each file has 1 line per detection and 5 values, bbox + category. Find more details about [the format here](https://github.com/AlexeyAB/Yolo_mark/issues/60#issuecomment-401854885).
- [x] Add similar export button on export modal as we have for Coco
- [x] .txt files and .jpeg files should in the same directory and with the same name (if options below is checked)
- [x] Add option to include in zip image files (the img have the same name as the .txt file)
- [x] Code function with input: export parameters from graphQL (dataset id, add images true/false)  Output: Zip File to export
  - [ ] SW adds object to cache
  - [ ] Use SW upload function to output the files
  - [ ] In YOLO resolver output url to the cache
- [Mockup](https://www.figma.com/file/4J9xyfF5v4B98MuKVO0Y8V/Labelflow-mockups?node-id=2482%3A12056)
image

**Question**
Should we go for YOLO Darknet TXT or YoloV5?

etiennedupont:
# Feature

## Work performed

- Created a generic export resolver
- Encapsulated the zipping and options dependant logic in the resolver for the existing coco

## Resolved issues

Closes #387


typescript/web/src/components/export-button/modal/__stories__/export-options-modal.tsx
```diff
@@ -1,12 +1,13 @@
 import { ExportOptionsModal } from "../export-options-modal";
 import { chakraDecorator } from "../../../../utils/chakra-decorator";
 import { apolloDecorator } from "../../../../utils/apollo-decorator";
+import { ExportFormat } from "../../../../../../graphql-types/src/graphql-types.generated";
```
RaynalHugo, line 4:
```suggestion
import { ExportFormat } from "@labelflow/graphql-types";
```

typescript/common-resolvers/src/export/index.ts
```diff
@@ -0,0 +1,53 @@
...
+const exportDataset = async (
+  _: any,
+  args: QueryExportDatasetArgs,
+  { repository }: Context
+) => {
+  const fileExport = await generateExportFile(args, { repository });
+  const outUrl = (await repository.upload.getUploadTargetHttp(uuidv4()))
+    ?.uploadUrl;
+  await repository.upload.put(outUrl, fileExport);
+  return outUrl;
```
RaynalHugo, line 46: Isn't there a more straightforward way to this when running in the Service Worker? Let's discuss this during the DSM.

typescript/web/src/components/export-button/modal/export-options-modal.tsx
```diff
@@ -47,20 +45,27 @@ const OptionLine = ({
...
+  const formatOptionsInformation =
+    formatsOptionsInformation[exportFormatLowerCase];
+  const optionsOfFormat = exportOptions[exportFormatLowerCase];
+  // const optionsOfFormat = exportOptions[exportFormatLowerCase];
+  // console.log(`
+  // exportOptions = ${JSON.stringify(exportOptions, null, 1)}
+  // exportOptions[exportFormatLowerCase] = ${JSON.stringify(exportOptions[exportFormatLowerCase], null, 1)}
+  // optionsOfFormat = ${JSON.stringify(optionsOfFormat, null, 1)}
+  // `);
   return (
```
RaynalHugo, line 69:
```suggestion
  const optionsOfFormat = exportOptions[exportFormatLowerCase];
  
  return (
```

typescript/web/src/components/export-button/modal/formats.ts
```diff
@@ -0,0 +1,71 @@
...
+    [optionKey in keyof ExportOptions[formatKey]]: {
+      title: string;
+      description: string;
+    };
+  };
+} = {
+  coco: {
+    exportImages: {
+      title: "Export image files",
+      description: "Zip images together with the annotation file",
```
RaynalHugo, line 48: Maybe we could add in the description a warning to the user saying that exporting images will take more time. 

(And do the same in `yolo.exportImages.description`)

typescript/web/src/connectors/resolvers/__tests__/export-to-coco.ts
```diff
@@ -133,7 +133,7 @@ const createLabelWithLabelClass = (imageId: string, labelClassId: string) => {
   });
 };
 
-describe("Exporting a dataset to coco format", () => {
+describe.skip("Exporting a dataset to coco format", () => {
```
RaynalHugo, line 136: It would be nice to reactivate those tests

typescript/web/src/components/export-button/modal/index.tsx
```diff
@@ -78,64 +68,27 @@ export const exportCocoDataset = async ({
...
+    data: { exportDataset: exportDatasetUrl },
+  } = await client.query({
+    query: exportQuery,
+    variables: {
+      datasetId,
+      format,
+      options: {
+        coco: { ...options.coco, name: datasetName },
+        yolo: { ...options.yolo, name: datasetName },
+      },
```
RaynalHugo, line 82: Nitpicking, feel free to ignore:

We could pass only the options of the chosen format, it would make the request a bit lighter.

typescript/web/src/components/export-button/modal/formats.ts
```diff
@@ -0,0 +1,71 @@
...
+      description: "Zip images together with the annotation directory",
+    },
+    avoidImageNameCollisions: {
+      title: "Avoid image names collision",
+      description:
+        "A unique identifier will be added to the image's name. Recommended when exporting the images",
+    },
+    includePolygons: {
+      title: "Include polygons",
+      description: "Replace polygons by their bounding box when exporting",
```
RaynalHugo, line 68: I think the title and the description are a bit contradicting themselves. 
According tot the title, I feel like I need to enable this to include polygons.
According to the description I feel like if I enable this I will not have polygons but bounding boxes.

crubier:
I have also one big remark: I would like to keep the directory structure like `typescript/common-resolvers/src/format-coco/export/index.ts`, and not like in this PR where it's like `typescript/common-resolvers/src/export/format-coco/index.ts`. 

The reason behind that is that soon we will have `import/format-coco` and `export/format-coco`, so the COCO-related code (for example) will be separated in two folders. I prefer to keep all the COCO related stuff inside the `format-coco` folder.  

Of couse this will be the same remark for other formats than COCO

typescript/web/src/connectors/resolvers/__tests__/export-to-coco.ts
etiennedupont: Finally I removed those tests, because the output of the resolvers is now a URL so it is more intricate to get the content of the file from that, and given that it's even more difficult when dealing with yolo format since it's a zip with several files. Core functions are tested in `typescript/common-resolvers/src/export/format-yolo/__tests__/export-to-yolo.ts` and `typescript/common-resolvers/src/export/format-coco/coco-core/__tests__/converters.ts`.

crubier (approved):

crubier:
@etiennedupont Ci blocked

PR #425, opened 2021-09-06
Fix Bug weird touch move on android on openlayers

crubier:
# Feature

## Work performed

Apply the fixes for this bug https://github.com/openlayers/openlayers/issues/10757

## Results

Much better labelling experience on android

## Problems encountered

None


typescript/web/src/pages/_app.tsx
```diff
@@ -64,6 +64,14 @@ const App = (props: AppProps & InitialProps) => {
...
             <ApolloProvider client={client}>
               <Meta />
+              <Head>
+                {/* Prevent unwanted scroll of app on mobile */}
+                {/* See https://stackoverflow.com/questions/10592411/disable-scrolling-in-all-mobile-devices */}
+                <meta
+                  name="viewport"
+                  content="width=device-width, initial-scale=1.0"
+                />
+              </Head>
```
nacho-carnicero, line 74: This does not prevent any scrolling no? It rather sets the appearance of the content on the first load right?

crubier: Yes, I put the comment, and then later I removed the bit that actually block the scroll because 1) this was not the real source of the problem and 2) this was blocking scroll on website. 

So yes we can change this comment

nacho-carnicero (approved):

PR #428, opened 2021-09-07
Add clarity tracking code

### LINKED ISSUE #426: Add clarity + cookie banner to website
**Tasks**
The goal is to be able to better understand the usage of the product by users.

- [ ] Add cookie acceptance banner to app & website
- [ ] Plugin https://clarity.microsoft.com/

crubier:
# Feature

## Work performed

Closes #426 

https://clarity.microsoft.com/projects/view/8cfef08lrk

## Problems encountered

Linking Clarity and Google Analytics (from the Clarity web App) crashes...

Screenshot 2021-09-07 at 17 54 39

Some WTF problems with chakra. Cookie consent banner swallows the rest of the app 

I managed to reproduce it. I think it’s caused by Next SSR. The server renders the cookie bar as the first div in the page, and then when react rehydrates on client side which has a cookie, react does NOT render the cookie bar, so the app gets put as the first div in the page, which is inside the cookie bar. It throws this error in console

Screenshot 2021-09-08 at 18 10 38

Screenshot 2021-09-08 at 17 23 53

Seems to be caused by

Screenshot 2021-09-08 at 18 25 40


crubier:
Need to talk about how the Telemetry / cookie consent should look like in the app. I suggest 

 1. adding it to the Welcome modal
 2. adding an option to toggle it on/of in the user preference menu
 
The traditional cookie banner can be done on the website.

nacho-carnicero (approved):

PR #439, opened 2021-09-10
Prefetch previous and next images

### LINKED ISSUE #188: Prefetch images to make image navigation faster
in order to make image navigation faster:

  - Prefetch the 1 next image details query in Graphql, and the 1 previous image
  - Prefetch the image content (fetch with the same headers as openlayers does) of the previous and next images
  - Prefetch the labels of the next and previous images

nacho-carnicero:
# Feature

## Work performed

- Prefetch previous and next images when being in the labelling platform
- Fix a bug where the welcome modal would try to create the tutorial dataset even if the user already had one

## Results

The images are properly fetched (there is no loading spinner when navigating from one image to the next one even if the user never visited that image) but this did not solve the glitch problem of images navigation.

## Resolved issues

Closes #188

### DIFF
--- a/typescript/web/src/components/labelling-tool/openlayers-map/openlayers-map.tsx
+++ b/typescript/web/src/components/labelling-tool/openlayers-map/openlayers-map.tsx
@@ -28,6 +28,7 @@ import {
   DrawingToolState,
 } from "../../../connectors/labelling-state";
 import { theme } from "../../../theme";
+import { useImagePrefecthing } from "../../../hooks/use-image-prefetching";
 
 const empty: any[] = [];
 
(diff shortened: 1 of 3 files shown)


crubier:
Do you see a concrete difference in perfomance ? I think that indeed, it does not change much. If no one sees a difference then maybe it's not worth merging this PR (But it was still worth trying it though)

RaynalHugo:
> Do you see a concrete difference in perfomance ? I think that indeed, it does not change much. If no one sees a difference then maybe it's not worth merging this PR (But it was still worth trying it though)

crubier (approved):

crubier:
> > Do you see a concrete difference in perfomance ? I think that indeed, it does not change much. If no one sees a difference then maybe it's not worth merging this PR (But it was still worth trying it though)
> 
> It avoids a loading spinner the first time you navigate between images. But it doesn't solve the resizing problem.

Ok indeed, cool

PR #440, opened 2021-09-10
Fix bug in change class shortcut on azerty keyboards

### LINKED ISSUE #380: Shortcuts `0`,`1`-`9` to set a class on a label do not work on azerty keyboard.
**Describe the bug**
Shortcuts `1`-`9` to set a class on a label do not work on azerty keyboard. Instead the direct keys below the numbers (`&`, `é`, `"`, `'` ...) work.

**To Reproduce**
1. Create a class
2. Right click on a label
3. On an Azerty keyboard, press `shift` + `1` to set the label to class 1. I does not work.
4. Press `&` (on Mac), it works.

**Expected behavior**
Add to the keys under the numbers, the numbers themselves as shortcuts.

**Screenshots**
IMG_5480

**Desktop (please complete the following information):**
 - OS: Mac 11.5.2
 - Browser Brave
 - Version 1.28.105

**Additional context**

nacho-carnicero:
# Feature

## Work performed

- Fix bug in change class shortcut on azerty keyboards
- Fix bug in keys listener in edit label class menu that missing a dependency

## Results

This PR makes sure that when pressing a number on AZERTY keyboards (so actually pressing `SHIFT`+ `key`) the class is properly changed.

change-class

## Caveats

Instead of adding the two possibilities (class is changed both when pressing `SHIFT`+`key` or just `key`) I decided to implement only the change when pressing properly a number. Otherwise we would have allowed to change classes by pressing `SHIFT`+`key` on non-AZERTY keyboards, which causes interferences with other shortcuts in some keyboards.

For example in a spanish keyboard the slash is made with `SHIFT`+`7`, so when pressing that key we would not know if the user wants to set the class corresponding to shortcut 7 or if he wants to use the `/` shortcut to focus the input field.

We could add a detector of keyboard and only allow to change classes by pressing `key` or SHIFT+`key` on AZERTY keyboards but I think it adds complexity for something that might not be worth it.

## Resolved issues

Closes #380

### DIFF
--- a/typescript/web/src/components/labelling-tool/openlayers-map/edit-label-class.tsx
+++ b/typescript/web/src/components/labelling-tool/openlayers-map/edit-label-class.tsx
@@ -75,9 +75,12 @@ export const EditLabelClass = forwardRef<
     [labelClasses, datasetId]
   );
   useHotkeys(
-    keymap.changeClass.key,
+    "*", // We have to manually check if the input corresponds to a change class key because otherwise on AZERTY keyboards we can't change classes when pressing numbers
     (keyboardEvent) => {
-      if (isOpen) {
+      if (
+        keymap.changeClass.key.split(",").includes(keyboardEvent.key) &&
+        isOpen
+      ) {
         // We do not want to interfere with the class menu shortcuts if this modal is closed
         const digit = Number(keyboardEvent.code[5]);
         const indexOfLabelClass = (digit + 9) % 10;
--- a/typescript/web/src/components/labelling-tool/options-tool-bar/edit-label-class-menu.tsx
+++ b/typescript/web/src/components/labelling-tool/options-tool-bar/edit-label-class-menu.tsx
@@ -151,7 +151,7 @@ export const EditLabelClassMenu = () => {
       }
     },
     {},
-    [labelClasses, onSelectedClassChange]
+    [labelClasses, onSelectedClassChange, isContextMenuOpen, setIsOpen]
   );
 
   return (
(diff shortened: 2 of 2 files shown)


typescript/web/src/components/labelling-tool/openlayers-map/edit-label-class.tsx
```diff
@@ -75,9 +75,12 @@ export const EditLabelClass = forwardRef<
...
+    "*", // We have to manually check if the input corresponds to a change class key because otherwise on AZERTY keyboards we can't change classes when pressing numbers
     (keyboardEvent) => {
-      if (isOpen) {
+      if (
+        keymap.changeClass.key.split(",").includes(keyboardEvent.key) &&
+        isOpen
+      ) {
         // We do not want to interfere with the class menu shortcuts if this modal is closed
         const digit = Number(keyboardEvent.code[5]);
         const indexOfLabelClass = (digit + 9) % 10;
```
crubier, line 86: You did not have to change this but it works on Azerty and Qwerty keyboards ?

nacho-carnicero: Yes check the GIF, I change between spanish and french keyboards. How would you have done it without changing this? I had a look at what had been done to support the slash hotkey and applied the same thing here

crubier: I would have used `keyboardEvent.key` instead of `keyboardEvent.code` , but it's fine if it works like that ✅

crubier (approved):

PR #445, opened 2021-09-13
Feature/responsive breadcrumbs

### LINKED ISSUE #422: Collapse breadcrumbs when too wide
**Description**
We want to avoid having breadcrumbs taking the navbar full width. Inspired from Google drive breadcrumbs UI:

- [ ] Set minimum spacing between breadcrumbs and tools to Xpix (**see screenshot 1**) @nicOwlas 
- [ ] Display a maximum of 4 breadcrumbs
- [ ] If breadcrumbs are larger than max width then shorten each breadcrumb (**see screenshot 2**). Question: what would be the logic?
- [ ] If there are more than 4 breadcrumbs (**screenshot 3**)
  - [ ] First one to the left is always visible
  - [ ] Last two are visible except on mobile (screen small)
  - [ ] The intermediate ones are replaced by `...`
  - [ ] When clicking on `...` a list displays the collapsed breadcrumbs

Note: to keep it simple, I would not take into account the case where the name of a dataset is very long for instance

### Screenshot 1
image

### Screenshot 2
image

### Screenshot 3
Sterblue

### Mockups
https://www.figma.com/file/4J9xyfF5v4B98MuKVO0Y8V/Labelflow-mockups?node-id=2586%3A15770
image

crubier:
# Feature

## Work performed

Closes #422 

- [x] Made a collapsed breadcrumbs components
- [x] Made a Breadcrumbs Component which uses the collapsed breadcrumbs component depending on the screen size
- [x] Use this Breadcrumbs Component  in all pages

## Results

Screenshot 2021-09-13 at 17 06 32

Screenshot 2021-09-13 at 21 40 13

Screenshot 2021-09-13 at 21 40 27

Screenshot 2021-09-13 at 21 40 34

Screenshot 2021-09-13 at 21 40 54

## Problems encountered

- Had a problem when putting breadcrumbs in a menu, had to use a dirty hack to fix it ([see code](https://github.com/labelflow/labelflow/pull/445/files#diff-65666d6aea3e2814b067a3180b91d407690492be59058052dc4e17abdbb3cfa9R68))
- Had to use another little trick because otherwise the ouline of the collapsed ellipsis button was hidden ([See code](https://github.com/labelflow/labelflow/pull/445/files#diff-a984a67572ef58c707a021b0c53fc16642684ef924e10b7e7f148bfab047591aR59) )

## Caveats

The switch to display X breadcrumbs and collapse the others is NOT based on the length of text in breadcrumbs, but is based on responsive breakpoints. So it's an heuristic. There are cases (see a screenshot above) where individual breadcrumbs are cut short with an elipsis.

## Resolved issues

Closes #422


nicOwlas:
Nice! Small remarks:
- I find the elements a bit tall in the menu, there is a lot of vertical spacing between the different breadcrumbs. Could give it a try to make it a bit more compact?
- The `>` icon in the menu is smaller than in the navbar
- There is more spacing between the icon `>` and the text in the menu than in the navbar
- Can the triple dot be a bit more heavy? I like this one https://icon-sets.iconify.design/ls/etc/
- When there are some ellipsis, is the full name displayed when the mouse hovers over an item?

crubier:
Solved in next commit

Screenshot 2021-09-14 at 11 15 59

nicOwlas:
Looking good!
Did you limit the width of the area the breadcrumbs can occupy? On a very large screen there is room to display more

image

crubier:
> Looking good!
> Did you limit the width of the area the breadcrumbs can occupy? On a very large screen there is room to display more
> 
> image

Yes I limit individual breadcrumbs (because you asked for it). The total width is only limited by the spacing with the buttons on the right of the topbar

nicOwlas:
I do not remember to have asked for this and I do not read it in the issue.

crubier:
> I do not remember to have asked for this and I do not read it in the issue.

I remember but whatever, I just made it 3x larger

RaynalHugo (approved):

PR #450, opened 2021-09-17
Fix bug of errors being thrown in service worker from create dataset modal

### LINKED ISSUE #436: Error thrown when typing to create a new dataset
**Describe the bug**
When creating a new dataset there is an error thrown each time the user types a new letter of the dataset name. This is not blocking anything for the user but pollutes our error reporting tool

**To Reproduce**
- Open the app
- Go to the datasets page
- Click on create dataset
- Type a dataset name and click on create
- Go to Sentry, you will see there are a bunch of errors thrown, one for each letter typed

**Expected behavior**
No errors should be thrown when typing the name of a non-existing dataset. This happens because we want to make sure the name of the dataset is unique (or it's slug) and therefore we check each time the user types something if there is a dataset with the same slug already. To do so we use the `dataset` resolver in `typescript/common-resolvers/src/dataset.ts` which throws when there is no dataset with a certain slug. We should therefore either modify the current resolver to not throw if there is no dataset with a certain slug, or create a specific resolver for this that does not throw.

**Tasks**

- [x] Create search resolver (blocked by #354 ) which does not return error when no datasets are return
- [x] Replace current query in `upsertModal` component with the `searchDataset` resolver

nacho-carnicero:
# Feature

## Work performed

- Add a query called `searchDataset` that allows to return `null` or `undefined` when querying a dataset that does not exist
- Call this query instead of the normal `dataset` query in `typescript/web/src/components/datasets/upsert-dataset-modal.tsx`
- Skip the `getDatasetByIdQuery` in `typescript/web/src/components/datasets/delete-dataset-modal.tsx` when `datasetId` is not defined to avoid throwing another error

## Results

No more errors are being thrown from the backend when creating a dataset

## Resolved issues

Closes #436

### DIFF
--- a/data/queries.graphql
+++ b/data/queries.graphql
@@ -27,6 +27,7 @@ type Query {
 
   dataset(where: DatasetWhereUniqueInput!): Dataset!
   datasets(where: DatasetWhereInput, first: Int, skip: Int): [Dataset!]!
+  searchDataset(where: DatasetWhereUniqueInput!): Dataset
 
   workspace(where: WorkspaceWhereUniqueInput!): Workspace!
   workspaces(first: Int, skip: Int): [Workspace!]!
@@ -37,10 +38,11 @@ type Query {
   user(where: UserWhereUniqueInput!): User!
   users(first: Int, skip: Int): [User!]!
 
-  exportDataset(where: ExportWhereUniqueInput!, 
-                format: ExportFormat!, 
-                options: ExportOptions
-                ): String!
+  exportDataset(
+    where: ExportWhereUniqueInput!
+    format: ExportFormat!
+    options: ExportOptions
+  ): String!
 
   debug: JSON!
 }
--- a/data/__generated__/schema.graphql
+++ b/data/__generated__/schema.graphql
@@ -293,6 +293,7 @@ type Query {
   labels(where: LabelWhereInput, first: Int, skip: Int): [Label!]!
   dataset(where: DatasetWhereUniqueInput!): Dataset!
   datasets(where: DatasetWhereInput, first: Int, skip: Int): [Dataset!]!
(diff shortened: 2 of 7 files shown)


nacho-carnicero:
⚠️ This is blocked by #354

data/queries.graphql
```diff
@@ -27,6 +27,7 @@ type Query {
 
   dataset(where: DatasetWhereUniqueInput!): Dataset!
   datasets(where: DatasetWhereInput, first: Int, skip: Int): [Dataset!]!
+  searchDataset(where: DatasetWhereUniqueInput!): Dataset
```
RaynalHugo, line 30: Usually, the signature of a search would be more something like 
`searchDataset(where: DatasetWhereInput!): [Dataset!]!`

I agree that for our use case this is not needed, but I think it might be better to make it more "standard". What do you think?

RaynalHugo:
I changed the target branch to the one of  #354 to make the review easier. Don't forget to change it back to main before merging (I think Github does it automatically if #354 is merged)

data/queries.graphql
crubier: Yes indeed

nacho-carnicero: Yes maybe it's a good thing to do, although in our case `DatasetWhereInput` is used to list the datasets of a workspace:

```graphql
input DatasetWhereInput {
  workspaceSlug: String!
}
```

So if we do this it complexifies a bit the queries using `DatasetWhereInput ` since you might want to filter by workspace + by name

crubier: Ah yes indeed it's a unique input so no need to have an array. but the "search" name is a bit misleading I guess.

crubier (approved):

PR #454, opened 2021-09-20
Add access control to dataset

### LINKED ISSUE #418: Implement access control based on workspace
**Tasks**

- [x] Tweak common resolvers to add the access control logic
  - [x] Workspace access control: check memberships
  - [x] Dataset access control: check if user has access to the workspace
  - [x] Image access control: check if user has access to the dataset
  - [x] Label class access control: check if user has access to the dataset
  - [x] Label access control: check if user has access to the image
  - [x] Membership access control
  - [x] User access control
- [x] For local entities, for Dexie repository: implement in the most trivial way: user can list every entity of his local workspace.
- [x] For online workspaces implement in prisma repository: forward the `where` condition to prisma getter and setter.
- [x] Add nested resolver to be able to query workspace from dataset
- [x] Add tests for access control in `typescript/db/src/resolvers/__tests__`

Blocked by #317

nacho-carnicero:
# Feature

## Work performed

- Added access control to entities (workspace, dataset, image, label class, label, membership, user)
- Added workspace nested resolver to dataset resolver to be able to query the workspace from the dataset

## Problems encountered

In order to add the workspace nested resolver I had to add the workspace operations to the repository (both Prisma and local ones)

## Caveats

- I added some tests for the dataset resolvers in the `db` folder in order to test the access control. This makes the tests on the dataset resolver to be split between the previous existing tests in `typescript/web/src/connectors/resolvers/__tests__/dataset.ts` and the new ones in `typescript/db/src/resolvers/__tests__/dataset.ts`

## Resolved issues

Closes #418


typescript/common-resolvers/src/types.ts
```diff
@@ -53,7 +58,15 @@ export type DbDataset = Omit<
   | "labelClasses"
   | "labelClassesAggregates"
   | "workspace"
->;
+> & { workspaceSlug: string };
+
+export type DbWorkspace = Omit<
+  GeneratedWorkspace,
+  "__export typename" | "type" | "datasets" | "memberships" | "plan"
```
RaynalHugo, line 65: is there a key called "__export typename" ? 🤯

typescript/db/schema.prisma
```diff
@@ -63,7 +63,7 @@ model Dataset {
   workspaceSlug String
   workspace     Workspace    @relation(fields: [workspaceSlug], references: [slug], onDelete: Cascade)
 
-  @@unique(fields: [workspaceSlug, slug], name: "workspaceSlugAndDatasetSlug")
+  @@unique(fields: [workspaceSlug, slug], name: "slugs")
```
RaynalHugo, line 66: Last time, @crubier asked to name this `workspaceSlugAndDatasetSlug` to be consistent with `workspaceSlugAndDatasetName`. Was he ok to do this change?

typescript/db/src/dev/apollo-server.ts
```diff
@@ -22,8 +21,22 @@ export const schemaWithResolvers = addResolversToSchema({
...
+        ? Object.fromEntries(
+            cookie.split(/; */).map((c) => {
+              const [key, ...v] = c.split("=");
+              return [key, decodeURIComponent(v.join("="))];
+            })
+          )
+        : {};
+    const session = await prisma.session.findUnique({
+      where: { sessionToken: parsedCookie?.["next-auth.session-token"] },
+    });
```
RaynalHugo, line 38: Isn't it what getSession does? 
https://next-auth.js.org/getting-started/client#getsession

data/dataset.graphql
```diff
@@ -5,7 +5,7 @@ input DatasetCreateInput {
 }
 
 input WorkspaceSlugAndDatasetSlug {
-  datasetSlug: String!
+  slug: String!
```
RaynalHugo, line 8: What motivated you to do this change?
You wanted to remember the good old times of our last PR when you spent 1 hour changing all the inputs of queries from `slug` to `datasetSlug` & `workspaceSlug` 🙃

RaynalHugo (commented):
You added Access Control on Datasets and Workspaces but I think we should add them on all entities, no?

With the current implementation, a user could list all images of all users, no?

data/dataset.graphql
nacho-carnicero: Believe me it was painful 😅 
I did this because the Prisma schema was defined as this:

```
model Dataset {
  id            String       @id @default(uuid()) @db.Uuid
  createdAt     DateTime     @default(now())
  updatedAt     DateTime     @updatedAt
  name          String
  slug          String
  images        Image[]
  labelClasses  LabelClass[]
  workspaceSlug String
  workspace     Workspace    @relation(fields: [workspaceSlug], references: [slug], onDelete: Cascade)

  @@unique(fields: [workspaceSlug, slug], name: "workspaceSlugAndDatasetSlug")
  @@unique(fields: [workspaceSlug, name], name: "workspaceSlugAndDatasetName")
}
```

As you can see the `unique` field relates `workspaceSlug` to `slug` so when I was trying to query a dataset from slugs I would get an error from Prisma saying that the variable `datasetSlug` was not known in the type `workspaceSlugAndDatasetSlug`. The other option would have been to do a `@map` in the Prisma schema from `slug` to `datasetSlug` for Dataset entities, but this would have changed all the query results for dataset entities. For instance, by doing the `@map` we would have had to query a dataset slug like this:

```gql
dataset{
  id
  datasetSlug
}
```

typescript/common-resolvers/src/types.ts
nacho-carnicero: Good catch! I'll change this

typescript/db/schema.prisma
nacho-carnicero: If we leave `workspaceSlugAndDatasetSlug ` then we can't do this when querying Prisma:

```gql
query dataset{
  dataset(where:{slugs:{slug:"my-first-dataset", workspaceSlug:"carnicero"}}){
    id
    name
  }
}
```

Instead, we have to do this:

```gql
query dataset{
  dataset(where:{workspaceSlugAndDatasetSlug:{slug:"my-first-dataset", workspaceSlug:"carnicero"}}){
    id
    name
  }
}
```

So I went ahead and changed the name since we agreed it was nicer in the API to have it this way. We can move it back to `workspaceSlugAndDatasetSlug` but it would mean that we need to change all the queries for dataset from slugs to the above query (even the local ones)

typescript/db/src/dev/apollo-server.ts
crubier: I agree, for everything related to auth, we should use next-auth hooks.

data/dataset.graphql
crubier: Hmm indeed, it feels like the is no good solution... 

Personally I would instead have renames Dataset.slug to Dataset.datasetSlug . But It's not ideal either...

typescript/db/src/dev/apollo-server.ts
nacho-carnicero: Yes but `getSession` would not work, probably because we are in a standalone server without NextJS 🤷

crubier: https://next-auth.js.org/getting-started/client#getsession

Did you try `const session = await getSession({ req })` ? I'm not sure if the req object is specific to NextJS

nacho-carnicero: Ok I think that now I remember why it would not work, from the docs it says:

>It calls /api/auth/session and returns a promise with a session object, or null if no session exists

Here we do not have any endpoint at that route, so it would not return any session.

crubier: Indeed, and I didnt have this in mind because it's new in Next Auth v4. Ok then I think your solution is good

nacho-carnicero:
> You added Access Control on Datasets and Workspaces but I think we should add them on all entities, no?
> 
> With the current implementation, a user could list all images of all users, no?

You're right, I thought there would be another issue for that since in #418 it was only mentioned to add it for dataset and workspace, but I do not see any other issue for access control for the other entities, I'll add it for them as well

nacho-carnicero:
Tests fixed ✅ @RaynalHugo can you have another look 🙏 ?

typescript/common-resolvers/src/types.ts
nacho-carnicero: Done in dcf831969de86dd10b1d5bcb12a4af1b5f5d5e33

RaynalHugo:
> Tests fixed ✅ @RaynalHugo can you have another look 🙏 ?

Will do!

typescript/db/src/repository/index.ts
```diff
@@ -1,62 +1,115 @@
...
       return createdLabel.id;
     },
-    count: countLabels,
-    delete: async ({ id }) => {
+    count: async (whereWithUser) => {
+      if (!whereWithUser) {
+        return 0;
+      }
+      const { user, ...where } = whereWithUser;
+      if ("datasetId" in where) {
```
RaynalHugo, line 83: Shouldn't we add something like `checkUserAccessToDataset` here?

typescript/db/src/repository/label.ts
```diff
@@ -2,30 +2,45 @@ import { PrismaClient, Prisma } from "@prisma/client";
...
-export const countLabels: Repository["label"]["count"] = async (where) => {
-  if (!where) {
+export const countLabels: Repository["label"]["count"] = async (
+  whereWithUser
+) => {
+  if (!whereWithUser) {
     return 0;
   }
+  const { user, ...where } = whereWithUser;
   if ("datasetId" in where) {
```
RaynalHugo, line 16: same here, shouldn't we check that the user has access to the dataset?

typescript/web/src/connectors/repository/label.ts
```diff
@@ -6,7 +6,12 @@ import { list } from "./utils/list";
...
 
-export const countLabels = async (where?: LabelWhereInput) => {
+export const countLabels = async (
+  whereWithUser?: LabelWhereInput & { user?: { id: string } }
+) => {
+  const { user, ...wherePossiblyEmpty } = whereWithUser ?? {};
+  const where =
+    Object.keys(wherePossiblyEmpty).length < 1 ? null : wherePossiblyEmpty;
   if (where) {
     if ("datasetId" in where) {
```
RaynalHugo, line 16: And same here, should we check if the user has access to the dataset?

typescript/db/src/repository/index.ts
nacho-carnicero: Yes you're totally right, I'll add it, nice catch!

typescript/web/src/connectors/repository/label.ts
nacho-carnicero: In this case we don't need it as we are in the local repository and we assume that the user has access to its local datasets

typescript/db/src/repository/index.ts
nacho-carnicero: Added in 928c65a8cc5638b940bd2f6023aa02b3b0caac42. Actually it shouldn't be needed because the only way we can list or count labels of a dataset is through a dataset query where we already test if the user has access to it. But I guess it does not hurt to add this

typescript/db/src/repository/label.ts
nacho-carnicero: Added in 928c65a8cc5638b940bd2f6023aa02b3b0caac42

RaynalHugo (approved):

crubier:
🎉

PR #472, opened 2021-09-28
Feature/split local workspace urls

RaynalHugo:
# Feature

## Work performed

- Implemented a switch between the local client and distant client depending on the page url
- updated all queries where `worskpaceSlug: "local"` was hardcoded.

## Results

🚨⚠️  This PR doesn't include the workspace switch ⚠️🚨

You can now have different workspaces depending on the URL you are visiting.

How to test this:
- run a PostgresQL db using `docker compose up -d` in `typescript/db`
- yarn the web app using `NEXT_PUBLIC_FEATURE_SIGNIN=true NEXT_PUBLIC_ENDPOINT=http://localhost:3000/api/graphql yarn dev` in `typescript/web`
- go to `localhost:3000/graphiql` 
- sign in using the button on the top bar
- create a workspace using a mutation in the playground and retrieve its slug
- navigate to `/SLUG_OF_YOUR_WORKSPACE` 
- check that everything works as it should.

## Problems encountered

- upload using drag and drop works on the remote server, however, pasting URLs doesn't. (I'll open an issue)

## Caveats

- On the _app, I cannot use `router.query.workspace` to switch between the local and the distant client. On the first render, `router.query` always equals to `{}`, we have to wait for `router.isReady` to be true. But, if we do so, it means that all the useQueries will be triggered on one client and then the client changes on the second render. To bypass this, I use `window.location`. This means that it won't work in SSR.

## Newly raised issues
https://github.com/labelflow/labelflow/issues/484


crubier:
Denied UX changes, some stories are broken

PR #476, opened 2021-09-30
Bug/#469 cannot rename dataset

### LINKED ISSUE #469: Cannot rename a dataset
**Describe the bug**
Cannot rename a dataset

**To Reproduce**

1. Create a dataset
2. Try to rename it
3. Error: `This name is already taken`

**Expected behavior**
Possibility to rename a dataset

**Tasks**

- [ ] Solve bug
- [ ] Add end-to-end tests to check this feature is not broken by future changes.

**Screenshots**
image

**Additional context**

etiennedupont:
# Feature

## Work performed

- Fixed bug when renaming dataset
- Added one suite of end-to-end tests for dataset creation, renaming, deletion

## Resolved issues

Closes #469

### DIFF
--- a/typescript/web/src/components/datasets/upsert-dataset-modal.tsx
+++ b/typescript/web/src/components/datasets/upsert-dataset-modal.tsx
@@ -136,7 +136,7 @@ export const UpsertDatasetModal = ({
 
   useEffect(() => {
     if (
-      existingDataset != null &&
+      existingDataset?.searchDataset != null &&
       !loadingExistingDatasets &&
       existingDataset?.searchDataset?.id !== datasetId &&
       variablesExistingDatasets?.slug === slugify(datasetName, { lower: true })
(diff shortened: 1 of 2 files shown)


typescript/web/src/components/datasets/upsert-dataset-modal.tsx
```diff
@@ -136,7 +136,7 @@ export const UpsertDatasetModal = ({
 
   useEffect(() => {
     if (
-      existingDataset != null &&
+      existingDataset?.searchDataset != null &&
```
nacho-carnicero, line 139: Maybe we should check here that `existingDataset?.searchDataset?.id != null` as it is the property that we are comparing to the current datased ID later on?

crubier:
```suggestion
      existingDataset?.searchDataset?.id != null &&
```

PR #488, opened 2021-10-04
Delete images

### LINKED ISSUE #130: Delete images in the project
**Tasks**
The mockup was designed to have a quick win and avoid redesign of the image component (which will be necessary when will have tags, selection etc.)
- [ ] In Images tab allow to delete by clicking on the bin. Display the confirmation modal.
- [x] Create Delete Image resolver
  - [x] Delete labels on the deleted image
  - [x] Delete image from storage

Mockup (check links for ground truth)

- https://www.figma.com/file/4J9xyfF5v4B98MuKVO0Y8V/Labelflow-mockups?node-id=2698%3A12087
- https://www.figma.com/file/4J9xyfF5v4B98MuKVO0Y8V/Labelflow-mockups?node-id=2703%3A12359

nacho-carnicero:
# Feature

## Work performed

- Allow to delete images in a dataset

## Results

delete-images

## Resolved issues

Closes #130


typescript/common-resolvers/src/image.ts
```diff
@@ -220,6 +221,29 @@ const createImage = async (
...
+  });
+  await Promise.all(
+    labelsToDelete.map((label) =>
+      repository.label.delete({ id: label.id }, user)
+    )
+  );
+  await repository.image.delete({ id: imageId }, user);
+  return imageToDelete;
+};
+
```
etiennedupont, line 246: Shouldn't you delete the image in the cache as well, calling `await repository.upload.delete(image.url)`?

etiennedupont (commented):
Maybe you could add some lines in the golden path to test image deletion?

typescript/common-resolvers/src/image.ts
crubier: I agree I think

nacho-carnicero: Done in 9ad999fbcf54906b60f0c3ae4fd0fa1fb414fb3c

As a side note I added a check to detect if the image that is being deleted is not one from the tutorial dataset

crubier (approved):

PR #512, opened 2021-10-14
Feature/create workspace popover

RaynalHugo:
# Feature


typescript/common-resolvers/src/types.ts
```diff
@@ -139,9 +140,12 @@ export type Repository = {
     update: Update<DbDataset, DatasetWhereUniqueInput>;
   };
   workspace: {
-    add: Add<WorkspaceCreateInput>;
+    add: Add<WorkspaceCreateInput & { slug: string }>;
```
crubier, line 143: ~~That does not look right at first.~~

~~Why do you not add `slug: string` to `WorkspaceCreateInput` directly instead ?~~

~~It seems weird to me that the resolver does not accept exactly the input type.~~

~~Also, being able to specify `slug` and `name` separately in the GraphQL query is a security problem. The name => slug conversion should be done inside the graphql resolver, otherwise user can abuse the API to create workspaces where slug does not match name.~~

typescript/db/src/repository/workspace.ts
```diff
@@ -24,18 +24,17 @@ export const addWorkspace: Repository["workspace"]["add"] = async (
   if (typeof user?.id !== "string") {
     throw new Error("Couldn't create workspace: No user id");
   }
-  const slug = slugify(workspace.name, { lower: true });
```
crubier, line 27: ~~As said above, I disagree with this change. The slugification should be done server side to avoid security problems.~~

crubier: Ok nevermind got it, it's not done client side, it's done in the resolver.

typescript/common-resolvers/src/types.ts
crubier: Ok got it, you moved this, not to the client, but to the resolver, to commonalize it. Good

typescript/db/src/resolvers/workspace.ts
```diff
@@ -64,10 +71,29 @@ const createWorkspace = async (
     );
   }
 
+  const slug = slugify(args.data.name, { lower: true });
```
crubier, line 74: Ahhh ok got it here !

**bruits d'éditage des commentaires precedents* *

PR #513, opened 2021-10-15
Add profile page

### LINKED ISSUE #507: Create account settings
### Tasks

- [ ] Add account settings page
  - [ ] Add `settings` as a reserved name that cannot be used by a workspace)
  - [ ] Url of the page: https://labelflow.ai/settings/profile
- [ ] Add `Settings` in user menu in the navigation bar
- [ ] Use [Form with shadow](https://pro.chakra-ui.com/components/application-ui/forms). with following edits
  - [ ] Only keep the Name&Avatar section
  - [ ] Remove the buttons change name and change avatar
  - [ ] Replace the name field by a text box with the name pre-filled
  - [ ] Add a Cancel / Save button at the bottom on the right hand side (labelflow-500 for the Save button)
  - [ ] No Breadcrumbs, only LabelFlow logo in the top left.
- [ ] ~~Add to user entity in DB a URL field (URL to the user picture)~~
- [ ] ~~Store the user picture on supabase~~
- [ ] ~~fill the avatar with the picture centered~~
- [ ] ~~If size limit is exceeded, display alert notification~~
- [ ] ~~Send verification email if the user updates his email address~~
- [ ] ~~Mark the email address as unverified until the user clicks in the email~~

### Questions / Remarks
- ~~Unless there is an easy solution, let's skip the modal to crop/reframe uploaded profile pictures~~
- ~~What limit size is reasonable for the picture?~~
- ~~For now, if a user has two accounts (GitHub + Email), if he enters his email address on his GitHub account, we will send a verification email and we would end up with 2 accounts for one email address which is problematic. Need to address merging of account rapidly.~~

### Mockup
[Form with shadow](https://pro.chakra-ui.com/components/application-ui/forms)

Screenshot 2021-10-14 at 11 07 38

nacho-carnicero:
# Feature

## Work performed

- Add profile page

## Results

profile

## Resolved issues

Closes #507

### DIFF
--- a/typescript/db/src/resolvers/__tests__/user.ts
+++ b/typescript/db/src/resolvers/__tests__/user.ts
@@ -384,7 +384,12 @@ describe("user query", () => {
     ).user.create({
       data: { id: testUser1Id, name: "test-user-1" },
     });
-    loggedInUser.id = testUser1Id;
+    await (
+      await getPrismaClient()
+    ).user.create({
+      data: { id: testUser2Id, name: "test-user-2" },
+    });
+    loggedInUser.id = testUser2Id;
     await expect(queryUser(testUser1Id)).rejects.toThrow(
       "User not authorized to access user"
     );


crubier:
> @nicOwlas could you check if you're happy with the result?
> The only thing that I did not put that was in the mockup was the "added at" field, we could do it but it would involve adding a new library to convert the ISO string format to a human readable format, and I don't think the added value for the user is big. But let me know if you want it and I'll add it

You can use date-fns for that

typescript/db/schema.prisma
```diff
@@ -153,296 +162,7 @@ model VerificationToken {
 }
 
 /////////////////////////////////////////////////////////////////////////////
-//////////////////////            Strapi          ///////////////////////////
```
nacho-carnicero, line 156: Why did you remove all the Strapi related schema?

typescript/web/src/components/settings/workspace/profile.tsx
```diff
@@ -0,0 +1,198 @@
...
+    };
+  }>(updateWorkspaceMutation, {
+    variables: {
+      name: workspaceName !== workspace?.name ? workspaceName : null,
+      image: null,
+      workspaceSlug: workspace?.slug,
+    },
+    refetchQueries: ["getWorkspaces"],
+    onCompleted: (createdData) => {
+      console.log(`Moving to /${createdData.updateWorkspace.slug}/settings`);
```
nacho-carnicero, line 73: Can be removed

typescript/web/src/components/settings/workspace/profile.tsx
```diff
@@ -0,0 +1,198 @@
...
+      router.push(`/${createdData.updateWorkspace.slug}/settings`);
+    },
+    onError: (caughtError: any) => {
+      if (caughtError instanceof ApolloError) {
+        toast({
+          title: "Needs to be signed in",
+          description:
+            "Only signed-in users can to create and share Workspaces online, please sign in.",
```
nacho-carnicero, line 81: Message not very clear

typescript/web/src/components/settings/workspace/profile.tsx
```diff
@@ -0,0 +1,198 @@
...
+  const [updateWorkspace, { error }] = useMutation<{
+    updateWorkspace: {
+      id: string;
+      name: string;
+      image?: string | null;
+      slug: string;
+    };
+  }>(updateWorkspaceMutation, {
+    variables: {
+      name: workspaceName !== workspace?.name ? workspaceName : null,
```
nacho-carnicero, line 67: Why do you do this check? It means that if at some point the save button is not disabled when `workspaceName === workspace?.name` then by clicking on save it will set the workspace name to `null`, not sure we want that

typescript/web/src/pages/[workspaceSlug]/settings/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+  const {
+    data: getWorkspaceDetailsData,
+    previousData: getWorkspaceDetailsPreviousData,
+  } = useQuery(getWorkspaceDetailsQuery, {
+    variables: { workspaceSlug },
+    skip: workspaceSlug == null,
+  });
+
+  const getWorkspaceDetailsFinalData =
```
Comment (nacho-carnicero) [inline typescript/web/src/pages/[workspaceSlug]/settings/index.tsx:38]: Why is this needed?

typescript/db/schema.prisma
crubier: Because I found out that it's possible ! It was always very annoying but I found a way to keep it in the DB but remove it from here so it's cool

typescript/web/src/components/settings/workspace/profile.tsx
crubier: True

nacho-carnicero:
LGTM!
I'll update the access control tests that is broken following my changes in the access control logic on User and we're good to go!

PR #535, opened 2021-10-25
End to end tests for online workspaces

### LINKED ISSUE #519: End to end tests for online workspaces
**Tasks**

- [ ] Add to GitHub actions (check what was done [in this PR](https://github.com/labelflow/labelflow/pull/457/files))
  - [ ] [GitHub Documentation](https://docs.github.com/en/actions/using-containerized-services/creating-postgresql-service-containers)
  - [ ] Create following tests: https://www.figma.com/file/56hJc9jl9oedCWIfxj7TyU/State-Diagram?node-id=23%3A695
- [ ] Create a test specific DB

nacho-carnicero:
# Feature

## Work performed

- Add possibility to execute tests against a postgresql database in end to end tests
- Use a separate DB for testing

## Caveats

Now we need to execute a different dev server to execute locally cypress tests. This can be done by executing `yarn dev:test` in `typescript/web` or `yarn dev:web:test` at the root of the repo.

There was also a change in how the environment variables are defined. Now we have 4 env vars files:
- `.env.production` (not committed) contains basically what we all had in `.env`, which is the secrets for production and the DB setup for prod
- `.env.local` (not committed) contains only the secrets needed to develop and test locally. This env var is loaded whenever executing tests or the dev server so that the secrets are accessible.
- `.env.development` (committed) contains the necessary information to setup the connection to the DB when developing, and contains as well all the env vars that are needed to the app (secrets are empty and will be overwritten by the env vars in `env.local`)
- `env.test` (committed) is the same as above but for testing

In order to setup your environment properly, once this is merged you should:
- rename your `.env` files to `.env.production`
- create a `.env.local` file at the root and put there the secrets you need to develop/test
- create a symlink from `.env.local` to an `.env.local` file in each folder needed (`/web`, `/infrastructure`, `/db`)

## Resolved issues

Closes #519


typescript/web/src/components/workspace-switcher/workspace-menu/workspace-selection-popover/workspace-selection-popover.tsx
```diff
@@ -143,81 +144,95 @@ export const WorkspaceSelectionPopover = ({
...
-                  pr="2"
-                >
-                  {inputValue && (
-                    <>
-                      <CloseCircleIcon
-                        fontSize="2xl"
-                        onClick={reset}
-                        cursor="pointer"
-                        color={closeCircleIconColor}
+      <Portal>
```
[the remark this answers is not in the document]
nacho-carnicero: Because there was a property set in the Breadcrumb component to hide the overflow on the X direction. This property was inherited by all its children and notably the workspace selection popover, and this was causing Cypress to think the popover was not visible when actually it was. So rendering it in a portal enables to avoid inheriting that prop on the popover and Cypress does not complain anymore

.github/workflows/ci.yml
```diff
@@ -160,6 +174,10 @@ jobs:
         with:
           path: /home/runner/.cache/Cypress
           key: ${{ runner.os }}-cypress-${{secrets.ACTIONS_CACHE_VERSION}}-${{ hashFiles('yarn.lock') }}
+      - name: generate prisma
+        run: cd typescript/db && yarn deploy && yarn generate
+        env:
+          POSTGRES_EXTERNAL_URL: ${{secrets.POSTGRES_EXTERNAL_URL}}
```
crubier, line 180: Why do we use a secret here in the e2e tests ? It should be hardcoded to use the local db, not the production db

cypress-ci.json
```diff
@@ -0,0 +1,9 @@
+{
```
crubier, line 1: Why do we need a different cypress for the CI ? Why not just run next on port 3001 in the CI too ?

cypress/integration/online-access.ts
```diff
@@ -0,0 +1,58 @@
+describe("Online workspaces access", () => {
```
crubier, line 1: Can you add comments for the main steps of the test ? Makes them more legible and maintainable

See cypress/integration/classification.ts

crubier (approved):
I tested it online and it seems ok, review ok too

cypress-ci.json
nacho-carnicero: Deleted in a7c55e9663e2c8038cecd288c13623b66123b621

.github/workflows/ci.yml
nacho-carnicero: Actually in our secrets in Github `POSTGRES_EXTERNAL_URL` is pointing to localhost, so this is indeed the local DB, should I change it to a hardcoded one?

crubier: Strange... It shouldn't, it should be the Supabase URL. So yes here, localhost is not a secret, it should just be hardcoded for the test.

cypress/integration/online-access.ts
nacho-carnicero: Done in 71c9ccfcc8aae75ec8100476df54a944132cece1

.github/workflows/ci.yml
nacho-carnicero: Ok will replace all references to that secret in the CI file by the hardcoded localhost host (it's set as well for the unit tests)

nacho-carnicero: Done in 61cba8ee89c11c9cd29d46d835ec56359033e7d3

PR #545, opened 2021-10-29
Feature/#518 mask iog

### LINKED ISSUE #518: Investigate: apply mask on image but on bounding box to highlight the object being iog
**Taks**
Investigate this point, best option to have an overlay?

- [x] Apply mask on image but on bounding box to highlight the object being iog. Investigate with openlayers.
- [x] Bounding box can be resized like a bounding box (no translation to not interfere with inside or outside point creation)

Options
  - [x] https://openlayers.org/en/latest/examples/layer-clipping-vector.html
  - [x] Polygon with a hole where the bounding box is

### IOG storage
- [x] Create a DB column: `smartToolInput`
- [x] Store the iog json input
 ```
{
  "iog": {
    "version": "v1",
    "input": {
      "id": ID!,
      "imageUrl": String,
      "x": Float,
      "y": Float,
      "width": Float,
      "height": Float,
      "pointsInside": [[Float!]],
      "pointsOutside": [[Float!]],
      "centerPoint": [Float!]
    }
  }
}

```

Links: https://github.com/labelflow/labelflow/issues/508#issuecomment-943240472

### PR DESCRIPTION
**Pull request by etiennedupont**
# Feature

## Resolved issues

Closes #518

### CODE REVIEW

File: typescript/web/src/components/labeling-tool/openlayers-map/labels.tsx [hunk 1]
```diff
@@ -54,7 +60,11 @@ export const Labels = ({
     variables: { imageId: imageId as string },
   });
   const selectedLabelId = useLabelingStore((state) => state.selectedLabelId);
+  const selectedTool = useLabelingStore((state) => state.selectedTool);
   const labels = data?.image?.labels ?? previousData?.image?.labels ?? [];
+  const selectedLabel = labels.filter(
```
crubier, line 65: use `array.find()` instead of `array.filter()[0]` . It's faster because it stops iterating as soon as it finds it.

crubier (approved):

PR #548, opened 2021-11-03
Feature/generate thumbnails

crubier:
# Feature

## Work performed

1. Generate thumbnails, was an adventure:
   - ⚪  ⚪ Start working on thumbnail generation
   - ⚪  🟡 Tried to use wasm vips on server: It works with a few tweaks (see comment below)
   - 🔴  🟡 Tried to use wasm vips on client: Works in window, does not work in service worker  => `SharedArrayBuffer is not defined`, not defined in service worker. 
   - 🟡  🟡 Moved to jimp on client: Works but not on storybook => `Unexpected error while loading ./web/src/components/auth-manager/signin-modal/__stories__/signin-modal.tsx: TypeError: _path.default.join is not a function` 
   - 🟡  🟢 Moved to jimp on server: Works  
   - 🔴  🟢 Moved to async import of jimp on client: works in storybook, does not work in web => `Cannot load sw.1.js`, importScript not work for some reason
   - 🟡  🟢 Moved back to sync import of jimp on client => back to the problem with storybook
   - 🟡  🟢  Updated storybook => does not solve the problem with storybook
   - 🟢  🟢 Fixed client side `import Jimp from "jimp/browser/lib/jimp";` 
2. Use thumbnails:
   - Add thumbnailXXUrl fields to GraphQL (See schema below)
   - Add graph resolver for thumbnail, that fallback to image Url if thumbnail is not present or not ready yet.  (This allows to connect to storage without Thumbnails, external images, or also to NOT have to wait for thumbnail generation when importing images)
3. Added a "Graphiql" tab in Workspace page, because it's useful and it was not discoverable otherwise
   
```
input ImageCreateInput {
  id: ID
  datasetId: ID!
  createdAt: DateTime
  name: String
  path: String
  mimetype: String
  height: Int
  width: Int
  file: Upload # Either file or url or externalUrl are needed
  url: String # Either file or url or externalUrl are needed
  externalUrl: String # Either file or url or externalUrl are needed
  thumbnail20Url: String
  thumbnail50Url: String
  thumbnail100Url: String
  thumbnail200Url: String
  thumbnail500Url: String
}
```

## Problems encountered

- [x] See adventures above ⏫ 
- [x] In the end, the only easy way to generate thumbnails was jimp, which does it in pure JS (Slow). So now the storybook that includes creating images time out. I was able to fix this by removing the need to wait on thumbnail generation when importing image (Falling back to image URL in the UIs if thumbnail not ready yet) 
- [x] While testing with large images, I found that once deployed on Vercel, we hit a "5MB upload limit" for images larger than 5Mb,  I didnt know about this limit before. `Request Entity Too Large FUNCTION_PAYLOAD_TOO_LARGE` See this ... https://vercel.com/support/articles/how-to-bypass-vercel-5mb-body-size-limit-serverless-functions. => This is not a bug of this PR, it's already a bug on feature/workspace, so I created an issue https://github.com/labelflow/labelflow/issues/576
- [ ] ~~In Vercel functions (API routes), if your return a response to the request, all pending promises are stopped. So the technique of returning the image creation without waiting for thumbnail generation does not work in vercel~~ NOT SURE ACTUALLY
- [x] Needed to handle cases where you import external Images and you don't want to generate thumbnails for those VS cases where you import external Images and you DO want to generate thumbnails for those on the local server.


crubier:
Image processing with wasm vips instead of jimp:

```typescript
// ES6 module
// @ts-ignore
import Vips from "wasm-vips";

import "isomorphic-fetch";

import Blob from "fetch-blob";
import { getThumbnailUrlFromImageUrl } from "@labelflow/common-resolvers/src/utils/thumbnail-url";

globalThis.Blob = Blob;

const vipsPromise = Vips();

const defaultMaxImageSizePixel: number = 60e6;
const maxImageSizePixel: { [mimetype: string]: number } = {
  "image/jpeg": 100e6,
  "image/png": 60e6,
};

// Map from Vips loaders to MIME types
// Done manually and verified by me
// See https://www.rubydoc.info/gems/ruby-vips/Vips/Image for the list of all image.xxload() methods
// See https://www.iana.org/assignments/media-types/media-types.xhtml for list of all mime types
const vipsFormats: { [key: string]: string } = {
  analyzeload: "image/analyze", // Not sure
  csvload: "text/csv",
  fitsload: "image/fits",
  gifload: "image/gif",
  heifload: "image/heif",
  jp2kload: "image/jp2",
  jpegload: "image/jpeg",
  jpegload_buffer: "image/jpeg",
  jxlload: "image/jxl",
  magickload: "image/bmp", // Not sure, but vips use magick to load bmp See https://www.libvips.org/API/8.6/VipsForeignSave.html
  magickload_buffer: "image/bmp", // Not sure, but vips use magick to load bmp See https://www.libvips.org/API/8.6/VipsForeignSave.html
  matload: "application/x-matlab-data",
  matrixload: "image/matrix", // Not sure
  niftiload: "image/nifti", // Not sure
  openexrload: "image/x-exr", // See https://lists.gnu.org/archive/html/openexr-devel/2014-05/msg00014.html
  openslideload: "image/openslide", // Not sure
  pdfload: "application/pdf",
  pngload: "image/png",
  pngload_buffer: "image/png",
  ppmload: "image/x-portable-anymap", // See https://fr.wikipedia.org/wiki/Portable_pixmap
  radload: "image/vnd.radiance",
  rawload: "image/raw", // Not sure, maybe not relevant
  svgload: "image/svg+xml",
  tiffload: "image/tiff",
  tiffload_buffer: "image/tiff",
  vipsload: "image/vips", // Not sure
  webpload: "image/webp",
  webpload_buffer: "image/webp",
};

const validateImageSize = ({
  width,
  height,
  mimetype,
}: {
  width: number;
  height: number;
  mimetype: string;
}): {
  width: number;
  height: number;
  mimetype: string;
} => {
  const imageSize = width * height;
  const maxImageSize: number =
    maxImageSizePixel?.[mimetype] ?? defaultMaxImageSizePixel;
  if (imageSize > maxImageSize) {
    throw new Error(`
    Image is too big! Dimensions are ${width} x ${height} = ${Math.round(
      imageSize * 1e-6
    )}Mpx while limit is ${Math.round(maxImageSize * 1e-6)}Mpx
    `);
  }
  return {
    width,
    height,
    mimetype,
  };
};

/**
 * Given a partial image, return a completed version of the image, probing it if necessary
 */
export const processImage = async (
  {
    width,
    height,
    mimetype,
    url,
  }: {
    width: number | null | undefined;
    height: number | null | undefined;
    mimetype: string | null | undefined;
    url: string;
  },
  getImage: (url: string) => Promise<ArrayBuffer>,
  putImage: (url: string, blob: Blob) => Promise<void>
): Promise<{
  width: number;
  height: number;
  mimetype: string;
}> => {
  const vipsObject = await vipsPromise;
  // The type comes from typescript/db/src/@types/wasm-vips.d.ts
  const VipsImage: typeof vips.Image = vipsObject.Image as typeof vips.Image;

  const buffer = await getImage(url);

  try {
    const vipsImage = VipsImage.newFromBuffer(buffer);

    const result = {
      width: vipsImage.width,
      height: vipsImage.height,
      mimetype: vipsFormats[vipsImage.getString("vips-loader")],
    };

    const vipsThumbnail20 = VipsImage.thumbnailBuffer(buffer, 20).writeToBuffer(
      ".jpg"
    );

    putImage(
      getThumbnailUrlFromImageUrl({ url, size: 20, extension: "jpeg" }),
      new Blob([vipsThumbnail20], { type: "image/jpeg" })
    );

    const vipsThumbnail50 = VipsImage.thumbnailBuffer(buffer, 50).writeToBuffer(
      ".jpg"
    );

    putImage(
      getThumbnailUrlFromImageUrl({ url, size: 50, extension: "jpeg" }),
      new Blob([vipsThumbnail50], { type: "image/jpeg" })
    );

    const vipsThumbnail100 = VipsImage.thumbnailBuffer(
      buffer,
      100
    ).writeToBuffer(".jpg");

    putImage(
      getThumbnailUrlFromImageUrl({ url, size: 100, extension: "jpeg" }),
      new Blob([vipsThumbnail100], { type: "image/jpeg" })
    );

    const vipsThumbnail200 = VipsImage.thumbnailBuffer(
      buffer,
      200
    ).writeToBuffer(".jpg");

    putImage(
      getThumbnailUrlFromImageUrl({ url, size: 200, extension: "jpeg" }),
      new Blob([vipsThumbnail200], { type: "image/jpeg" })
    );

    const vipsThumbnail500 = VipsImage.thumbnailBuffer(
      buffer,
      500
    ).writeToBuffer(".jpg");

    putImage(
      getThumbnailUrlFromImageUrl({ url, size: 500, extension: "jpeg" }),
      new Blob([vipsThumbnail500], { type: "image/jpeg" })
    );

    return validateImageSize({
      width: width ?? result.width,
      height: height ?? result.height,
      mimetype: mimetype ?? result.mimetype,
    });
  } catch (e) {
    console.error(e);
    throw new Error(
      `Could not probe the external image at url ${url} it may be damaged or corrupted.`
    );
  }
};
```

crubier (commented):
self reviewed

crubier:
Tested online, it works:

- ✅ localhost, local workspace
- ✅ localhost, online workspace
- ✅ vercel, local workspace
- ✅ vercel, online workspace

Issue #575, opened 2021-11-12
Say something if user tries to login by email without filling email


crubier:
When the users clicks "login with email without filling the form (email adress), it should warn them beforehand, not say "mail sent"

<img width="1176" alt="Screenshot 2021-11-12 at 17 08 20" src="https://user-images.githubusercontent.com/2954572/141506593-bbea82c6-f7d6-4828-90b8-fda3fb6b7e5f.png">

RaynalHugo:
Done in #581

PR #581, opened 2021-11-15
Feature/validate email address on signin modal

RaynalHugo:
# Feature

## Work performed

Add email validation on sign-in modal

## Results

- Button is disabled when empty
image

- Button is enabled and focus is blue when valid
image

- Border is grey when valid and unfocused
image

- Button is disabled and border is red (when focused and unfocused) when invalid email (and not empty)
image

## Problems encountered

Chakra doesn't really let you use native `:valid` and `:invalid` CSS selectors without re-implementing all chakra styling. So, we have to keep a react state to handle this :/

## Resolved issues

#575

### DIFF
--- a/typescript/web/src/components/auth-manager/signin-modal/signin-modal.tsx
+++ b/typescript/web/src/components/auth-manager/signin-modal/signin-modal.tsx
@@ -23,6 +23,7 @@ import { FaGithub, FaGoogle } from "react-icons/fa";
 import { DividerWithText } from "./divider-with-text";
 import { Logo } from "../../logo";
 import { Features } from "./features";
+import { validateEmail } from "../../../utils/validate-email";
 
 const errors = {
   Signin: "Try signing in with a different account.",
@@ -57,6 +58,8 @@ export const SigninModal = ({
   setLinkSent: (email: string) => void;
 }) => {
   const [sendingLink, setSendingLink] = useState(false);
+  const [isEmailInvalid, setIsEmailInvalid] =
+    useState<boolean | undefined>(undefined);
 
   const performSignIn = useCallback(
     async (method, options = {}) => {
@@ -140,6 +143,9 @@ export const SigninModal = ({
 
               <DividerWithText>or sign in with email</DividerWithText>
               <form
+                css={{
+                  ":invalid": { button: { color: "red" } },
+                }}
                 onSubmit={(e) => {
                   e.preventDefault();
                   const email = (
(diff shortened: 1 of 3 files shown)


typescript/web/src/components/auth-manager/signin-modal/signin-modal.tsx
```diff
@@ -140,6 +143,9 @@ export const SigninModal = ({
 
               <DividerWithText>or sign in with email</DividerWithText>
               <form
+                css={{
+                  ":invalid": { button: { color: "red" } },
```
nacho-carnicero, line 147: Is this used for something?

typescript/web/src/components/auth-manager/signin-modal/signin-modal.tsx
```diff
@@ -161,8 +167,25 @@ export const SigninModal = ({
...
-                      <FormControl id="email">
+                      <FormControl
+                        id="email"
+                        isRequired
+                        isInvalid={isEmailInvalid}
+                      >
                         <Input
+                          onChange={(event) => {
+                            setIsEmailInvalid(
+                              !validateEmail(event.target.value)
```
nacho-carnicero, line 178: Maybe here we should use this:

```
!validateEmail(event.target.value) || event.target.value.length === 0
```

This poses a problem because when the user clicks on the form and it is empty it has a grey border. Then the user types something and then erases it, the border is red until the input is blurred. We therefore have here two different states of the border color for the same state of the input value, which can be confusing

typescript/web/src/components/auth-manager/signin-modal/signin-modal.tsx
```diff
@@ -161,8 +167,25 @@ export const SigninModal = ({
...
+                            );
+                          }}
+                          onBlur={(event) => {
+                            if (event.target.value.length === 0) {
```
nacho-carnicero, line 182: If you add the check on `event.target.value.length === 0` above then I don't think you need this

typescript/web/src/components/auth-manager/signin-modal/signin-modal.tsx
RaynalHugo: Oh no, good catch. This was from a -failed- attempt to use native CSS selector to manage the `:valid`/`:invalid` state of the form

typescript/web/src/components/auth-manager/signin-modal/signin-modal.tsx
RaynalHugo: I am not sure that the change you propose here does what you want.

Basically, you want to always make the border red when the length is 0, right?
`onChange` isn't called until you type the first character. Therefore, when clicking on the input, it will be blue until you type something. Then if you erase all its content it will become red. 

I think the best workflow would be:
- onFocus, make it blue if it contains a valid email, else, if invalid or empty => red
- onChange, valid => blue; invalid or empty => red
- onBlur, invalid => remains red, valid or empty => default state (grey)

Whatt do you think?

RaynalHugo: I implemented this behaviour in 578d9a11101cc610d74009425dd2fe78470496e5, if you want to give it a try

typescript/web/src/components/auth-manager/signin-modal/signin-modal.tsx
RaynalHugo: It is still needed otherwise the border always remains red after being blurred out. I think we want it to go back to the default state when it's empty

typescript/web/src/components/auth-manager/signin-modal/signin-modal.tsx
RaynalHugo: Changed discussed during he DSM have been done in aef5493a66000431fc052b9dd0e39fb8d4d5ed45

nacho-carnicero (approved):

PR #587, opened 2021-11-16
Fix bug when trying to upload many files

### LINKED ISSUE #580: Uploading many light images fails
**Describe the bug**
Uploading many images fails (every image is less than 1MB)

**To Reproduce**
Upload those images in an online dataset: https://drive.google.com/drive/folders/1JBGYWHVfe3PpTrekYVUNLPA3eC6F6cnV?usp=sharing
Screenshot 2021-11-15 at 11 46 02

- [ ] `getDatasets` fails: 
```
{"errors":[{"message":"Error querying the database: db error: FATAL: remaining connection slots are reserved for non-replication superuser connections","locations":[{"line":2,"column":3}],"path":["datasets"]}],"data":null}
```
- [ ] `createImageMutation` fails:
```
{"errors":[{"message":"User not authenticated","locations":[{"line":2,"column":3}],"path":["createImage"]}],"data":{"createImage":null}}
```
- [ ] `getDatasetsData` fails
```
{"errors":[{"message":"User not authenticated","locations":[{"line":2,"column":3}],"path":["dataset"]}],"data":null}
```
**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

nacho-carnicero:
# Feature

## Work performed

- Change database connection URL to connect to pgbouncer endpoint rather than connecting directly to the database
- Fix a maximum concurrency of 10 on the upload of images

## Results

Screenshot 2021-11-17 at 17 13 08

## Problems encountered

The upload of images concurrency has to be limited because PGBouncer has a maximum number of connections allowed as input (which defaults to 100, see `max_client_conn ` in the [pgbouncer docs](https://www.pgbouncer.org/config.html)). When executing the uploads from a serverless environment (such as Vercel) there is one connection made to the DB per image upload, therefore when uploading more than 100 images the `max_client_conn` is reached and the process fails. Since there is more than 1 query made to the DB per image upload, we need to lower the amount of images that can be concurrently uploaded. In this PR we limit it to 10, which is a value that works while not giving a too bad performance. 

## Resolved issues

Closes #580


nacho-carnicero:
There could be sometimes some problems with this, for instance when multiple users upload a lot of images at the same time. The solution for this I guess would be to be able to increase the amount of max connections in pgbouncer, or to have several DB instances. I have asked a question in the Supabase repo to see recommendations from the community for this:
https://github.com/supabase/supabase/discussions/3928

etiennedupont (approved):
Looks good to me 👌

RaynalHugo (approved):

Issue #582, opened 2021-11-16
Wrong error handling when trying to create an online workspace with an already existing slug (or name)


nacho-carnicero:
**Describe the bug**
When trying to create an online workspace with a name that is already taken we get the sign in popup as if the user was not signed in. Instead we should not allow creating an online workspace if the name is already taken, and display an explicit error to the user.

**To Reproduce**
1. Sign in
2. Click on create new workspace
3. Create a workspace with a slug that is taken
4. The sign in popup shows up, if you check the network tab of the dev tools you'll see the real error

**Expected behavior**
We should not allow creating an online workspace if the name is already taken, and display an explicit error to the user.

**Screenshots**
<img width="1680" alt="Screenshot 2021-11-15 at 18 12 41" src="https://user-images.githubusercontent.com/17384901/141965724-f45d3a6a-43fb-4aee-bd49-a6274680f421.png">

Issue #551, opened 2021-11-04
Cleanup code


crubier:
- [ ] Upgrade prisma from v2.30.0 to v3.3.0
- [ ] Rename "upload" repository and routes to "storage"
- [ ] Others ?

RaynalHugo:
## DX Proposal: Factorize errors

Instead of doing `throw new Error("user isn't logged in)` everywhere and with a slightly different error message sometimes. We could have our custom Error classes.
```typescript
import { UnauthenticatedUser } from "@labelflow/errors";

const resolver = (_, _, {user}) => {
  if (!user.id) {
    throw UnauthenticatedUser()
  }
}
```

We could enforce message template/formatting, add error codes and even write doc about errors.

I don't think that's critical to do it. But I think it's a good thing for newcomers or external contributors. And the sooner we start this, the easier it is.

RaynalHugo:
Another proposal: Investigate page redirections using Next middleware.

bbenoist:
Here are some suggestions:

* Split `data` schemas into 3 distinct parts `service-worker`, `online`, and `common`.

* Remove the `python/api` project if it's dead code.

* Don't group projects by tech, use a flat-structure (e.g. `projects/*`) and explicitly declare build targets and dependencies in their respective package/build system.

* Use meaningful file names and extensions. Examples:
  * instead of `__tests__/index.tsx`, use `my-component.test.tsx`
  * instead of `__stories__/index.tsx`, use `my-component.stories.tsx`
  * See https://github.com/labelflow/labelflow/tree/feature/delete-workspace/typescript/web/src/components/workspace-name-input for more examples

* Don't hardcode colors, reuse chakra contextual names or declare new ones (e.g.: `error` instead of `red.500` or `#ff0000`).

bbenoist:
Split `typescript/react-openlayers-fiber/src/renderer.ts` (1012 lines) and `typescript/react-openlayers-fiber/src/catalogue.ts` (764 lines)

bbenoist:
Remove commented-out dead code

RaynalHugo:
Cleanup `tsconfig.json` files (see https://labelflow.slack.com/archives/C022PCC0H0W/p1640167922172300)

We use symlinks instead of extends. It creates strange behaviours like missing `strict` mode on some packages`

bbenoist:
Evaluate https://codeclimate.com/ (free for OSS) or similar

PR #597, opened 2021-11-19
create dedicated isWorkspaceSlugAlreadyTaken resolver and use it on t…

RaynalHugo:
# Feature

## Work performed

- Created a new query resolver named `isWorkspaceSlugAlreadyTaken`. It returns a boolean. The purpose is to avoid making the entire workspaces public.
- Swapped the query used in the workspace creation modal and replaced it with the newly created one.

## Results

Now, you get a proper warning which prevents you from creating a workspace with an already taken slug.

However, it doesn't solve the original bug which was that you get asked to sign in if the query fails.
I think this should be done in a deeper refactor regarding error handling in general. See https://github.com/labelflow/labelflow/issues/551#issuecomment-960700298

## Problems encountered

Permissions were not really clearly defined here. I implemented what I think is the safest solution but you can now know if a workspace name already exists. 

## Resolved issues

#582

## Newly raised issues

Not new, but relates to https://github.com/labelflow/labelflow/issues/551#issuecomment-960700298

### DIFF
--- a/data/__generated__/schema.graphql
+++ b/data/__generated__/schema.graphql
@@ -378,6 +378,7 @@ type Query {
   searchDataset(where: DatasetWhereUniqueInput!): Dataset
   workspace(where: WorkspaceWhereUniqueInput!): Workspace!
   workspaces(first: Int, skip: Int, where: WorkspaceWhereInput): [Workspace!]!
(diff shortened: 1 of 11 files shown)


nacho-carnicero (approved):

nacho-carnicero:
There's a small UI issue, when the name of the workspace is protected, the border is not red anymore:
Screenshot 2021-11-23 at 12 00 49

Issue #601, opened 2021-11-22
Enable to test image uploads on end to end tests in CI


nacho-carnicero:
We should enable to test the image uploads on online workspaces in the end to end tests. In #571 we were going to use the Supabase docker-compose file to deploy in the CI an environment similar to the current backend we use, but using the filesystem locally as the storage provider instead of a bucket in a remote provider. With #599 we are going to switch the storage provider to S3, the idea is therefore to find an alternative that we can deploy in the CI so that we use the filesystem as a storage provider to be able to test image uploads. [Minio](https://min.io/) should be a good solution for this.

**Describe the solution you'd like (import visuals when possible)**
- [x] Deploy a minio instance in the docker compose used for the end to end tests
- [x] Use that minio instance as the endpoint of the storage provider for tests
- [ ] Add a basic image upload test to online workspace to the end to end tests to test that this works as expected
- [ ] Used the end point env var and token given by #599
```
accessKeyId: process.env?.LABELFLOW_AWS_ACCESS_KEY_ID!,
secretAccessKey: process.env?.LABELFLOW_AWS_SECRET_ACCESS_KEY!
```
Blocked by #599

Issue #576, opened 2021-11-12
Upload of files > 5Mb not supported on vercel


crubier:
Once deployed on Vercel, we hit a "5MB upload limit" which I didnt know about.... 

`Request Entity Too Large FUNCTION_PAYLOAD_TOO_LARGE` 

See https://vercel.com/support/articles/how-to-bypass-vercel-5mb-body-size-limit-serverless-functions. 

I think the solution would be to make the upload go directly to supabase, but I dont think it's possible for uploads... Probably need to open an issue on supabase: Need to be able to create a SignedUrl for UPload. 

<img width="1499" alt="Screenshot 2021-11-12 at 17 06 38" src="https://user-images.githubusercontent.com/2954572/141507173-5c87bdea-df59-415e-98f3-38ee00f3d966.png">

Blocked by #607

etiennedupont:
I started investigating the issue. Here are the options I envision:
- Wait for an upload signedUrl on Supabase as Vincent proposed, since it doesn’t exist yet => the easiest alternative if we're not in a hurry
- Use the Supabase client, client-side. We'd need to give extra care about security, and the fact we don’t use Supabase auth is not helping. Some references to do that: https://supabase.io/docs/guides/auth#row-level-security & https://github.com/supabase/supabase/tree/master/examples/nextjs-ts-user-management#postgres-row-level-security
- Use another storage method (Google Cloud, AWS) that supports upload SignedUrl, in an end-to-end fashion instead of Supabase => Probably the best option if we want to move fast
- Use another storage method (Google Cloud, AWS) as an intermediate. E.g. upload on a GC bucket then download it from the API serverless function and finally upload on Supabase using the client => makes things more intricate

RaynalHugo:
> * Wait for an upload signedUrl on Supabase as Vincent proposed, since it doesn’t exist yet => the easiest alternative if we're not in a hurry

Oh I was sure it existed but I've just seen that it's only for download, and not for upload ...

etiennedupont:
Issue created on Supabase
https://github.com/supabase/storage-api/issues/81

etiennedupont:
Fixed in #607

PR #607, opened 2021-11-23
Feature/aws connector

### LINKED ISSUE #599: Store Images on AWS
## Limitations with current implementation
- Vercel limits file upload to 5MB
- No Supabase CI on Mac

## Options
- Change from NextAuth to Supabase for authentification
  - Need to change Sign in page
  - Need to write RLS rules
- Change storage provider (AWS)
  - Easier

## Tasks option: Change storage provider (AWS)
- [ ] #626
- [x] Create 1 bucket `labelflow-images` on Sterblue AWS
- [x] Organize files similarly than on Supabase (thumbnails and original images)
- [x] Upload connector: test `uploadS3`

## Improvements
- [ ] #631

etiennedupont:
# Feature

## Work performed

- Added a `/dowloads/ ` API route to `GET` from s3, similar to supabase `/uploads/` counterpart (but without a `PUT` method)
- Refactored the existing s3 connector for upload
- Changed the key in storage from `datasetId/imageId.jpeg` to `workspaceId/datasetId/imageId.jpeg`
 
## Problems encountered

- There isn't full compatibility / symmetry with the Supabase way of storing:
   - Client-side, the `PUT` is apparently compulsory to be done as multipart (didn't find any alternative), while it's done regularly for s3
   - The API route that was added doesn't support `PUT`, meaning that the upload targets are signed URLs directly pointing to s3. This allows to fix the [5Mb issue ](https://github.com/labelflow/labelflow/issues/576)
   - I couldn't make the `repository.upload.put` for Supabase looking like a regular `fetch`, it still uses Supabase client. Normally this should be possible but I couldn't make it work, which seems to be linked to the fact that the upload needs to be multipart
-  Linked to the Minio issue #601 and work by @RaynalHugo : we can't currently use minio in tests with the aws SDK v3 that we use because of this issue https://github.com/minio/minio/issues/13779. The best option is probably to downgrade aws.
- To make tests work, since I couldn't use Minio,  I had to mock S3 client and `getSignedUrl` util and also `fetch`. I commented some line called `fetch.disableFetchMocks();`. I don't know the implications of that @nacho-carnicero.

## Resolved issues

Closes #599


typescript/common-resolvers/src/image.ts
```diff
@@ -125,7 +125,7 @@ export const getImageEntityFromMutationArgs = async (
       );
     }
 
-    await repository.upload.put(uploadTarget.uploadUrl, blob);
```
vincent-lecrubier-skydio, line 128: I really think you should not do that this way. 
If you want, you can change the resolver for uploadTarget to return the right URL.
But having the tool do an upload to the download URL is a recipe for future disaster, it makes no sense.

vincent-lecrubier-skydio (changes requested):

typescript/common-resolvers/src/image.ts
etiennedupont: Yes we talked about that, I changed the parameters from being an URL to the key in the storage directly

crubier: I'm not sure of the implications of doing that, but try to make sure that your solution will work well when we'll be able to plug several storage backends per dataset for example.

I feel that by having only the key be explicit, you make implicit assumption about the storage backend which is used, and now these assumptions will have to be shared somehow between the backend and the frontend.|

Will it work well if we say "This dataset is stored on the customer's S3 account in spexific bucket XYZ" ?

etiennedupont: The change came from the acknowledgement that the URL that was passed was systematically pruned to just keep the key. By digging a bit into that, there seems to be a semantic issue here: what would be an "uploadUrl" for s3? Is it the URL that points to the object? Then it's rather a download URL and cf your comment above, and you can't actually "put" to that URL. If it's a signedUrl for upload, then this function is trivial, which is okay, but it also already shows incompatibility with Supabase storage.

etiennedupont: I agree that the key is not sufficient to make the difference between let's say buckets / organizations inside a storage method, but there are even more parameters needed if we don't use signed URLs, like the API keys

crubier: We could (should) do for upload just like we do for download: serve a redirection to a signedUrl that is generated during « getUploadTarget » using the S3 credentials. But that URL being a public presigned URL does not require credentials, temporary credentials are actually encoded in the presigned URL, that’s the point of those.

Of course this is not compatible with supabase for now, but for supabase backed datasets you just keep the current behavior and we have that 5mb limitation. 

URLs are a universal way to represent a resource, they give us flexibility for the future. But keys seems to couple things which rubs me in the wrong direction

etiennedupont: I'm totally fine with using signedUrls, but then wouldn't it just remove the need for a `repository.get` / `repository.put` since they'll both simply become a disguised `fetch` ?

crubier: Yea indeed and that would be a good thing. BUT as of today for example out of the top of my head I know we’d still need repository.put in any case because to do a fetch put, you need an endpoint to do it, but due of the absence of upload presigned url in supabase, there are no endpoint to do fetch put to upload. So we created our own, but we can’t call our own endpoint from our own server side functions (e.g. to upload images contained in a zip file import). So we still need repository.put for that use case for example.

crubier: If the interface to upload and download could be a simple external url instead of a function call, for all storage providers, then we could do without repository.put and repository.get. That would be nice. But for now we are blocked because of this supabase thing.

etiennedupont: Yes clearly, if we want to keep supporting Supabase we'll need these function; at least as long as Supabase don't implement upload signed URLs.

etiennedupont: After discussing with the others during the DSM, we found out that the main blocker for removing those `repository.upload.get` / `repository.upload.put` lies on the local service worker side rather than for the supabase solution. Actually we can query our API routes from server-side functions, but the service worker can't call query itself.

So I'll revert to expecting URLs in `repository.upload.get` / `repository.upload.put` that really make sense at least for server-side, meaning you do `put` and `get` on those URLs. About serving a redirection for the upload part, we discussed that as well, I thought that would be the best option since it preserves symmetry with respect to downloads but unfortunately I think we'd bump again in the 5Mb limit (cc @nacho-carnicero) since we'd send over the image in the body when putting on the API route, even if we don't actually use that body. I'm not exactly sure about how redirection works, but to my mind it's unlikely that the route ignores the body, so we'd rather provide an upload signedUrl directly when calling `getUploadTarget`.

crubier: Yes, true, it's the SW the most problematic for this I forgot 👍 

For the 5Mb limit for uploads, I would try it just to be absolutely sure, but yes it makes sense.

nicOwlas:
Action @nicOwlas : test uploading large files

nicOwlas:
@etiennedupont the app does not start
image

typescript/web/src/components/import-button/import-images-modal/modal-dropzone/modal-dropzone.tsx
```diff
@@ -212,11 +233,18 @@ export const ImportImagesModalDropzone = ({
                 });
               }
 
-              const form = new FormData();
-              form.append("image", acceptedFile.file);
+              // SUPABASE upload
```
nacho-carnicero, line 236: Can this comment be removed?

.env.development
```diff
@@ -15,8 +15,9 @@ GITHUB_CLIENT_SECRET=""
 GOOGLE_CLIENT_ID=""
 GOOGLE_CLIENT_SECRET=""
 EMAIL_SERVER=""
-SUPABASE_API_URL=""
```
nacho-carnicero, line 18: Should we maybe leave the Supabase env vars as an example of the vars needed to use the Supabase client?

typescript/common-resolvers/src/export/index.ts
```diff
@@ -48,7 +48,7 @@ const exportDataset = async (
       origin
     )
   )?.uploadUrl;
-  await repository.upload.put(outUrl, fileExport);
+  await repository.upload.put(outUrl, fileExport, req);
```
nacho-carnicero, line 51: I can't see any put function in any of the three repos that needs `req`, why was this added?

typescript/db/src/repository/upload-s3.ts
```diff
@@ -1,64 +1,102 @@
...
-//   const signedUrl = await getSignedUrl(s3Client, command, {
-//     expiresIn: 3600,
-//   });
-//   return {
-//     __typename: "UploadTargetHttp",
-//     uploadUrl: signedUrl,
-//     downloadUrl: `${location}/${key}`,
-//   };
-// };
+export const uploadsRoute = "/api/downloads";
```
nacho-carnicero, line 33: This is only used for download so I think it'd be better to rename it to `downloadsRoute`

typescript/web/src/components/import-button/import-images-modal/modal-dropzone/modal-dropzone.tsx
etiennedupont: This is the only instance of the practical asymmetry between Supabase and S3 storage, so I'd prefer to keep it until we definitely discard Supabase for storage.

.env.development
etiennedupont: Of course, I don't know why I removed it

typescript/common-resolvers/src/export/index.ts
etiennedupont: Yes, that's a remnant of me trying to "fetch" regularly for Supabase: you needed to add the credentials in the request since they are checked when uploading on the upload route. I agree that I don't use them but in the end I felt like it could be useful in the future plus it's more symmetric with `repository.upload.get` that needs `req` for the same reason. What do you think?

etiennedupont:
> @etiennedupont the app does not start image

@nicOwlas I can't reproduce, it works on my side

nicOwlas:
> @nicOwlas I can't reproduce, it works on my side on the last deployment

All good, it works for me now 🤷

nicOwlas:
@etiennedupont: Erreur quand j’essaye d’importer le zip sur un online workspace dans la branche S3.

Import sur:
https://labelflow-5uavx58mb-labelflow.vercel.app/demo/datasets/powergrid/images?modal-import
image

etiennedupont:
Ok I'll have a look

nicOwlas:
It's working well with regular images larger than 5MB 🎉

nicOwlas:
> Ok I'll have a look

In fact I also have problems in local: 
image

typescript/common-resolvers/src/export/index.ts
nacho-carnicero: Ok let's leave it then, I agree it could be useful in the future

PR #608, opened 2021-11-23
Add Minio for development

RaynalHugo:
# Feature

## Work performed

- Added Minio in the DB docker compose file in order to emulate an S3 storage locally
- Added default en vars for development for `MINIO_ROOT_USER` and `MINIO_ROOT_PASSWORD`
- Added Minio in the CI on the jest and cypress tests in anticipation of future needs

## Results

You can try it locally. 
You need to have the AWS CLI installed.  

```sh 
cd typescript/db
# Start Minio (storage is the name of the service, it only starts minio, not postgres)
docker compose --env-file .env.development up -d storage
# Create a bucket
AWS_SECRET_ACCESS_KEY=minioadmin AWS_ACCESS_KEY_ID=minioadmin aws --endpoint-url http://127.0.0.1:9000/ s3 mb s3://testbucket
# list all bucket
AWS_SECRET_ACCESS_KEY=minioadmin AWS_ACCESS_KEY_ID=minioadmin aws --endpoint-url http://127.0.0.1:9000/ s3 ls  
```

You can also visit http://localhost:9001 to access Minio's console. Use `minioadmin` both as user and password.

## Problems encountered

- In the CI, I couldn't use the GitHub action Service API.  
There is no way currently with this API to specify the command to run when starting the docker image. To spin up minio, we need to run `docker run minio/minio server /data`. We don't have a way to specify `server /data` in GitHub.  
Also, minio doesn't offer a docker image with this predefined command (They do have it [here](https://github.com/minio/minio/blob/master/Dockerfile.cicd) but they don't host it on their docker hub, we could host it on our own, but we would need to keep it updated). 
- The AWS cli works, but I didn't tested with the JS SDK. @nacho-carnicero @etiennedupont it would be interesting to test it with the s3 connector you've just implemented.

## Caveats

⚠️ 🚨 Minio is under AGPL License 🚨 ⚠️ 
We can safely use it on our CI. But I am not sure we can communicate commercially on some on-premise development using minio. We need to investigate further.

## Resolved issues

#601


.github/workflows/ci.yml
```diff
@@ -159,7 +178,26 @@ jobs:
...
+      # And, as of today, there are no easy way to pass the `server /data` part.
+      # Therefore, we start minio in docker manually
+      - name: Setup storage with minio
+        run: |
+          docker run -d \
+                     --name minio \
+                     -p 9000:9000 \
+                     -e MINIO_ROOT_USER=$MINIO_ACCESS_KEY \
+                     -e MINIO_ROOT_PASSWORD=$MINIO_SECRET_KEY \
+                     minio/minio server /data
```
etiennedupont, line 200: I don't really know the implications of not being able to use Github Services, but if that's an issue, couldn't we thinly wrap this image with one applying those hardcoded run arguments?

RaynalHugo: I developed a bit on this point in the PR description. 
To be honest, I don't think there are huge differences when using this approach.
Minio did that [here](https://github.com/minio/minio/blob/master/Dockerfile.cicd). But they don't host it on docker hub ... We could host it on our side. Be it means we would have to maintain it up to date. 
And again, I don't know if it would change anything for GitHub actions

etiennedupont: Yes okay that's what I thought too!

PR #615, opened 2021-11-26
Experiment/on premise deployment

### LINKED ISSUE #566: On premise deployment - CI Publishes docker image
- [ ] The CI publishes one docker image called labelflow to docker hub.

Depends on #565

### LINKED ISSUE #567: On premise deployment - Labelflow docker compose
- [ ] Labelflow docker compose: supabase docker compose & labelflow docker

Depends on #566

### LINKED ISSUE #686: Minimize the size of docker image for on-premise deployment

RaynalHugo:
# Feature

## Work performed

Creating a PR to avoid forgetting this branch:
WIP of a single root docker compose to start the entire infra.

## Results

- Starts a db and applies migrations (but migrations aren't awaited to start the web server)
- IOG server is started but the dev server doesn't connect to it yet
- Minio Storage is added but the dev server doesn't yet use it (still linked to supabase)


bbenoist:
# Work done

* Make Docker configuration build and run correctly on my machine
* Find a solution to allow contacting MinIO from another Docker container using the hostname `localhost` instead of the Docker internal hostname `minio`
* Make Auto-Polygon work with local workspaces
* Minor code cleanups

## `NEXT_PUBLIC_*` environment variables propagation

As with other React static-site generation tools, the `NEXT_PUBLIC_*` environment variables are only propagated to the source code at build-time ([source](https://nextjs.org/docs/basic-features/environment-variables#exposing-environment-variables-to-the-browser)).

I had to move them as Docker build arguments so that the URL could be reused by the app.

As-is, this forces us to build a specific Docker image each-time we want to use a different value for `NEXT_PUBLIC_VERCEL_URL` and `NEXT_PUBLIC_IOG_API_ENDPOINT`.

We must implement a way to load these variables on the server-side and forward them to the web-browser when needed.

### Recommendations

* [ ] For `NEXT_PUBLIC_IOG_API_ENDPOINT`, use server-side props or create a `iogUrl` query
* [ ] Don't reference `NEXT_PUBLIC_VERCEL_URL` anywhere in the code.

## MinIO on a local Docker development environment

MinIO requests have to be made from both the web browser and the Docker container (as we cannot share our S3 credentials publicly and should avoid passing by our API servers to upload files).

On the other hand, we must use a single URL during the whole process since the URL signing features of the S3 API (and MinIO) don't allow to generate any URL with a hostname different than the one used to make the request.

In most use-cases, this is not a problem since the MinIO or S3 service is expected to be reachable anywhere with the same hostname.

But, when using Docker on a local machine for development or testing purposes, you'd need to reach MinIO with the hostname `localhost` instead of `minio` as defined in our `docker-compose.yml` file.

As explained above, we must a single hostname reachable from anywhere in the stack.

Solution are:

a. Make the web-browser use the `minio` hostname (e.g. by editing the `/etc/hosts` file).

b. [Proxy `http://localhost:9000` (MinIO URL) requests sent from the `web` container to `http://minio:9000` (MinIO URL) with a reverse-proxy running within the container.](https://github.com/labelflow/labelflow/blob/experiment/on-premise-deployment/typescript/web/Dockerfile#L67)

Since solution **b** provides a zero-setup configuration when working on a local machine with Docker, we've decided to implement it.

Solution **a** is still available anyway as it requires no changes in our repository.

### Recommendations

Work is already done! See [my comments in the `Dockerfile`](https://github.com/labelflow/labelflow/blob/experiment/on-premise-deployment/typescript/web/Dockerfile#L67) for details.

I don't really like to use a reverse-proxy in such a manner but that's the cleanest way we found to avoid extra-configuration on the developer-side and keep our stack working seamlessly with a single `docker compose up` command.

To minimize the impact of the solution and don't deploy it on production images, I've used the Docker multi-phase build features to create a dedicated image which inherits from the production version and adds the reverse-proxy.

Usage of this image can be enabled by specifying the `local-dev` target as in the root `docker-compose.yml` file (might be moved in a `docker-compose.dev.yml` at some point).

## Usefulness of the Prisma migration Docker image

While I think that it's a good idea to have a Docker job which can execute the Prisma DB migration, I'm not sure whether we really need to have a dedicated Docker image for it.
Maybe that it's preferable to reuse the `web` image and give it a specific command.

This way we could:

* Keep providing a single Docker image for the main app (without Auto-Polygon)

* Reduce the storage required for the Docker images (not sure about this one since I've shared the install phase between both images).

For example, we could declare both the `ENTRYPOINT` and `CMD` in the `web` image:

```dockerfile
ENTRYPOINT ["/app/custom-entry-point"]
CMD ["start"]
```

`start` would be the default task when calling the `docker run` command without any argument and we could trigger the migration by passing `migrate` as custom command when starting the migration jobs.

### Recommendations

* [ ] Use a single image and entrypoint for both the `start` and `migrate` tasks.

## Drawbacks with using Docker a monorepo

Using a monorepo is (mostly) awesome!
But when it comes to building Docker images, a few problems becomes more painful than usual:

* Source code and development-only dependencies must not be installed in the final image.
* You should avoid rebuilding the common internal packages multiple times.
* You might need to manage multiple Docker files requiring different `.dockerignore` wildcards or build contexts.

After working on multiple projects with the same issue, I ended-up using the [Rush deployment features](https://rushjs.io/pages/maintainer/deploying/) which helps a lot with this task by creating minimal deployment bundles.

I've not looked at more yarn-compatible alternatives yet but I think that describing it might help the discussion.

Apart from other useful monorepo features, one of the key-advantages [Rush](https://rushjs.io/) is aware of both external and cross-project dependencies.  
This makes it possible to tell the tool to build and deploy only what's needed for a specific target.
When deploying, it automatically adjusts the dev dependencies and build packages which you can reuse outside of the source directory from where it has been built.

With such a tool, we could separate the Docker phases into (1 layer per-phase):

1. Copy package configuration files
2. Install every dependencies.
3. Copy the source code
4. Build the desired targets
5. Package build results
6. Deploy build results in separate images (1 per target)
7. Declare the `EXPOSE`, `ENTRYPOINT` or `CMD` entries and other target-specific stuff

### Recommendations

* [ ] Separate `package.json` copies and package installs in a first layer so that the dependencies install phase can be skipped when possible.

  _Unless creating a dedicated image (with its own build context and `.dockerignore` only allowing `package.json` and `yarn` files), your only solution will be to add a `COPY` instruction for each file needed by this step (since Docker cannot respect the source directory structure when copying files with wildcards such as `**/package.json`)._

* [ ] Minimize Docker images sizes and build time by adding more entries into the main `.dockerignore` file so that the `COPY . .` line won't trigger useless rebuilds and embed unneeded files.

* [ ] Investigate solutions such as [Rush deployment features](https://rushjs.io/pages/maintainer/deploying/) to minimize our maintenance costs and simplify the deployment process.

## The suffering of working with Docker on my MacBook Pro

While working on this subject, my computer was struggling to build the images.

* The build process is very slow and requires a lot of resources!  
  Although having 16GB RAM, my *2018* MacBook Pro was struggling.
  Sometimes, I could see the letters I'm typing appearing every 2 seconds one-by-one.
* I couldn't work on other tasks in parallel since:
  * Tests of other projects seems to be relying on external resources (db?, etc).
  * The DB schema in this branch was not the same than the other branch as wanted to work on.
  * My machine was struggling as explained above.
* The MacBook screen was flickering

### Recommendations

* [ ] Get new machines :trollface:
* [ ] Make unit-tests work without any external resource
* [ ] If the computer is on its knees, use [GitHub Codespaces](https://github.com/features/codespaces) (or another cloud-based Docker build solution) when working on Docker images

bbenoist:
# Work done

My primary goal for today was to minimize image size and build time.

As explained in [my previous debrief](https://github.com/labelflow/labelflow/pull/615#issuecomment-996526596), one way to minimize package sizes it to deployment tools such as [`rush deploy`](https://rushjs.io/pages/maintainer/deploying/).

In that sense, I've tried to see if there was issues or technical things which would prevent us to use such tools.

Here are the issues I've found:

* #685

* #683

* https://github.com/prisma/prisma/issues/6899 (also applies to NeXt)

RaynalHugo:
Closing this PR as it's not longer under active development

bbenoist:
Please wait Hugo 🙏

RaynalHugo:
Sorry, I thought this had gone stale. 
The maniac in me wanted to clean the list of my open PRs.

Issue #594, opened 2021-11-18
Auto polygon improvements


nicOwlas:
**Describe the bug**
Tested on: https://labelflow-git-feature-508-iog-workflows-labelflow.vercel.app/local/datasets

**To Reproduce**

- [x] Rename the tool to `Auto Polygon` instead of `Auto Polygon Tool`
- [x] Make the donut grey darker: [mockup](https://www.figma.com/file/4J9xyfF5v4B98MuKVO0Y8V/LabelFlow-mockups?node-id=3271%3A31271) (grey-700 with a pass through of 70% on Figma)
- [x] Remove the segmented control buttons (polygon/auto polygon) when in creation mode
https://user-images.githubusercontent.com/16015322/142401711-54636a00-71c4-445f-a068-19c3db35a063.mov
- [x] Do not display a border on the inside hole of the donut: [mockup](https://www.figma.com/file/4J9xyfF5v4B98MuKVO0Y8V/LabelFlow-mockups?node-id=3271%3A31271)
- [x] The first remove point cannot be deleted
- [x] Undo is buggy, should have the same behaviour as Polygon.
- [x] IOG doesn't work on online workspaces
- [x] Investigate if canvas is resized before inference

etiennedupont:
>  Do not display a border on the inside hole of the donut: mockup

Are you sure? This makes the resizing of the canvas harder to discover

etiennedupont:
Done in #624

PR #624, opened 2021-11-29
Feature/#594 auto polygon improvements

etiennedupont:
# Feature

## Resolved issues

#594


typescript/common-resolvers/src/iog-inference.ts
```diff
@@ -79,52 +130,114 @@ const runIog = async (
...
-      y,
-      height,
-      width,
-    };
-    await repository.label.update({ id: args.data.id }, newLabelEntity, user);
-  }
+  const { xInit, yInit } = {
+    xInit: Math.min(image.width, Math.max(0, args.data.x)),
+    yInit: Math.min(image.height, Math.max(0, args.data.y)),
+  };
```
RaynalHugo, line 188:
```suggestion
  const xInit = Math.min(image.width, Math.max(0, args.data.x)),
  const yInit = Math.min(image.height, Math.max(0, args.data.y)),
```

I find it easier to read this way, what do you think?

typescript/common-resolvers/src/iog-inference.ts
```diff
@@ -79,52 +130,114 @@ const runIog = async (
...
+    []
+  );
+  const [x, y, X, Y] = [
+    Math.min(...xCoordinates),
+    Math.min(...yCoordinates),
+    Math.max(...xCoordinates),
+    Math.max(...yCoordinates),
+  ];
+  const width = X - x;
+  const height = Y - y;
```
RaynalHugo, line 159: Nitpicking:
I feel like this could be further optimised. We could do this in a single loop/reduce.

But feel free to ignore this comment if you don't feel like it would change much.

typescript/web/src/components/labeling-tool/openlayers-map/labels.tsx
```diff
@@ -148,29 +148,29 @@ export const Labels = ({
...
-                    geometry: (feature) => {
-                      const coordinates = (feature as Feature<Polygon>)
-                        .getGeometry()
-                        .getCoordinates()[1];
-                      return new MultiPoint(coordinates);
-                    },
+                    // stroke: new Stroke({
+                    //   color: `${iogMaskColor}FF`,
+                    //   width: 2,
+                    // }),
```
RaynalHugo, line 156:
```suggestion
```

typescript/web/src/components/labeling-tool/openlayers-map/labels.tsx
```diff
@@ -148,29 +148,29 @@ export const Labels = ({
...
+                  //     }),
+                  //   }),
+                  //   geometry: (feature) => {
+                  //     const coordinates = (feature as Feature<Polygon>)
+                  //       .getGeometry()
+                  //       .getCoordinates()[1];
+                  //     return new MultiPoint(coordinates);
+                  //   },
+                  //   zIndex: 2,
+                  // }),
```
RaynalHugo, line 173:
```suggestion
                  }),
```

typescript/common-resolvers/src/iog-inference.ts
etiennedupont: Done!

etiennedupont:
@RaynalHugo I made the requested changes, can you have a quick second look?

nicOwlas:
@etiennedupont the sw fails to register on Brave and Firefox, can you reproduce?

## Brave
image
image
image

## Firefox
image

etiennedupont:
@nicOwlas can you try again on the last deployment?

nicOwlas:
@etiennedupont My functional feedbacks:
- Works well locally, I could remove the last outside point, edit it switching to select etc. 👍 
- Did you find out if the canvas is being resized before inference?
- I could not run iog on an online workspace. Error 400.
```
{errors: [{message: "The column `smartToolInput` does not exist in the current database.",…}],…}
```
- Could you display the modal alerting that "auto polygon processed images are analyzed on our server" also when the user activates the tool through its shortcut `a`
- Could you add a link to the documentation in the modal? [Link](https://labelflow.gitbook.io/labelflow/labelling-interface/label-types/auto-polygons)
image
- The image is greyed out outside the bounding box only after the inference is returned. Can the outside of the bounding box be greyed out as soon as the bounding box is drawn?
- I still see some issues with the undo. Like here in this recording, I add remove point, I undo them but the polygon is not the same as it was with the same points. In the recording below, the word boxes is well labeled without any outside points. After adding some outside points and undoing them I do not have the same polygon geometry as I was having initially.
https://user-images.githubusercontent.com/16015322/145974823-8f60f514-076c-44e1-a5e5-3c7f4189c48b.mov

nicOwlas:
Not linked to iog but worth noting in case a new bug is being introduced: I cannot delete an image in a dataset on this release while I can on the workspace release.

nicOwlas:
Another bug: I cannot upload an image in the local workspace
image

etiennedupont:
@nicOwlas could you make another check? Thanks!

etiennedupont:
@nicOwlas can you confirm that the bug you mentioned is fixed?

nicOwlas:
- No problem anymore on the local workspace 👍 
- Online workspace: I got this error when trying to add the first inside point (the subsequent inside points worked though and  I not able to reproduce in another browser)
image
Untitled 2

etiennedupont:
Thank you @nicOwlas. I tried on several pictures and I can't reproduce this error. Can you reproduce on this particular picture?

nicOwlas:
@etiennedupont good for me 👍  🎉

PR #636, opened 2021-12-06
Create a new prod DB

nacho-carnicero:
# Feature

## Work performed

- Create a new project in Supabase
- Migrate data from old project to new project
- Created a script to clone the old storage bucket to the new project
- Cleaned data:
  - Remove unused tables
  - Removed test data (test workspaces, datasets etc...)
  - Moved strapi related tables to a specific `strapi` schema
- Added a field to configure Strapi to connect to the right DB schema
- Modified the env vars in Heroku (for Strapi) and Vercel (only for production) to point to the new DB


typescript/scripts/src/copy-supabase-bucket.ts
```diff
@@ -0,0 +1,122 @@
+import { createClient, SupabaseClient } from "@supabase/supabase-js";
+import Bluebird from "bluebird";
```
RaynalHugo, line 2: Maybe it could be nice to give a bit of context for why this script exists on the repository.
A few lines of comments and a link to this issue/PR could do the job. 

I think it's important to mention that this script isn't used on a regular basis, it was more a one-shot.

bbenoist: @nacho-carnicero @RaynalHugo It's dead code then?

nacho-carnicero: I'd say it's not dead code, rather it's rather a script that was done to execute a punctual action that we might do again in the future (so I thought we would like to save it)

nacho-carnicero: Added comment in b47cd1d055e6e9b7dec0c24c321094438ddb1232

RaynalHugo: 👌

RaynalHugo (approved):

Issue #626, opened 2021-11-30
Downgrade aws s3 sdk to make Minio work on tests


etiennedupont:
Problem detailled in https://github.com/labelflow/labelflow/pull/607

Blocked by #607

PR #638, opened 2021-12-06
Enhancement/make s3 client compatible with minio

RaynalHugo:
# Feature

## Work performed

- Didn't downgrade the AWS JS SDK version. It was a potential solution listed in the issue, but there is another workaround.
- Use alternative endpoint API on AWS S3 JS SDK to make it compatible with Minio (See https://github.com/aws/aws-sdk-js-v3/issues/1941#issuecomment-824194714)
- Added corresponding dev env vars
- Updated db's docker-compose file to start Minio with the "labelflow" bucket already created

## Results

Now, when you start the backend using `yarn start` in `typescript/db`, it also spins up a Minio Server storing files on your system. If you use the default env vars when you start the web app, it should automatically connect to Minio to store uploaded images.

## Problems encountered

- AWS SDK doesn't work as expected and forces us to use a workaround https://github.com/minio/minio/issues/13779

## Caveats

We create 2 different S3 Clients and the `/api/downloads/` route is specifically hardcoded to work with S3. Maybe we should make this more generic if we want to keep the initial mindset of having different storages

## Resolved issues

closes https://github.com/labelflow/labelflow/issues/626


RaynalHugo:
> Great solution! I think you can now get rid of `typescript/db/src/resolvers/__tests__/storage.ts` that was there temporarily. Also, I suggest that you try to remove s3 client mocks in `typescript/db/src/resolvers/__tests__/image.ts` and `typescript/db/src/resolvers/__tests__/label.ts`, normally they shouldn't be needed anymore

I've managed to remove the s3 client mocks and it works. However, if I disable the mocks of `fetch` it doesn't work :/

RaynalHugo:
> > Great solution! I think you can now get rid of `typescript/db/src/resolvers/__tests__/storage.ts` that was there temporarily. Also, I suggest that you try to remove s3 client mocks in `typescript/db/src/resolvers/__tests__/image.ts` and `typescript/db/src/resolvers/__tests__/label.ts`, normally they shouldn't be needed anymore
> 
> I've managed to remove the s3 client mocks and it works. However, if I disable the mocks of `fetch` it doesn't work :/

@etienne, I've finally managed to remove those mocks. Can you have another look? :)

.env.test
```diff
@@ -10,4 +10,4 @@ EMAIL_FROM="LabelFlow <contact@labelflow.ai>"
 LABELFLOW_AWS_ACCESS_KEY_ID=minioadmin
 LABELFLOW_AWS_SECRET_ACCESS_KEY=minioadmin
 LABELFLOW_AWS_REGION=eu-west-1
-LABELFLOW_AWS_ENDPOINT=localhost:9000
\ No newline at end of file
+LABELFLOW_AWS_ENDPOINT=http://localhost:9000
```
nacho-carnicero, line 13: Don't we need here as well all the other env vars you added in `.env.development`?
```
MINIO_ROOT_USER=minioadmin
MINIO_ROOT_PASSWORD=minioadmin
LABELFLOW_AWS_ACCESS_KEY_ID=minioadmin
LABELFLOW_AWS_SECRET_ACCESS_KEY=minioadmin
LABELFLOW_AWS_REGION=eu-west-1
LABELFLOW_AWS_ENDPOINT=http://localhost:9000 
```

RaynalHugo: Hum good question
Classic "on my machine, it works" but to be honest, now that you say it, I don't understand why it works ahah.
I'll investigate.

I know that the correct env vars are set on the docker-compose file, but, still it should ask need the credentials while testing

nacho-carnicero: Are they set in your `env.local` maybe?

RaynalHugo: Ah no, I found out why. I didn't do the change in this PR, because they were already on the `.env.test` file (look at the lines above your comments). 
Only `MINIO_ROOT_USER=minioadmin` & `MINIO_ROOT_PASSWORD=minioadmin` are missing, but they are only needed if you want to start a minio server (so on the docker-compose). To dialog with the server, we only need the AWS credentials as we use the AWS SDK.

nacho-carnicero: Ah yes it makes sense! Thanks for the explanation

nacho-carnicero:
LGTM but I'll let @etiennedupont approve the changes when he feels it since you wanted him to have a look as well

etiennedupont (approved):

PR #616, opened 2021-11-26
Bug/fix facebook article preview

RaynalHugo:
# Feature

## Work performed

- Added Facebook app id. It seems to display the correct preview now.
- Update twitter SEO params to point to LabelFlow's one instead of Vincent's one.

## Results

https://developers.facebook.com/tools/debug/?q=https%3A%2F%2Flabelflow-git-bug-fix-facebook-article-preview-labelflow.vercel.app%2Fposts%2Fai-metrics-image-classification

image

Cute cat BTW 😻 

## Resolved issues

closes #612

PR #641, opened 2021-12-07
Set correct value for `og:url` of articles

RaynalHugo:
# Feature

## Work performed

Fixed image preview on Facebook. 

The fix on https://github.com/labelflow/labelflow/pull/616 was only working for Vercel preview deployments, it wasn't working on prod.

To fix this, I had to specify a correct value for `og:url`. 
By default, it was always pointing to `labelflow.ai`, from what I've read (I'm not the most experienced on this part) it has o point toward the URL of the article itself.

I also added a few meta tags specific to articles on the `/posts/[slug]` url.

## Results

Now I hope it works as expected, it should, but to make sure this is working, we have to deploy it on labelflow.ai

## Problems encountered

The behaviour was different between branch deployments and prod because we put a default `og:url` value to `labelflow.ai`, as explained on Facebook documentation [here](https://developers.facebook.com/docs/sharing/webmasters/getting-started/versioned-link/), cross domains references are ignored unless they have been approved on the admin dashboard.

### DIFF
--- a/typescript/web/src/pages/posts/[slug].tsx
+++ b/typescript/web/src/pages/posts/[slug].tsx
@@ -16,13 +16,9 @@ import { PostTitle } from "../../components/website/Blog/PostTitle";
 import "github-markdown-css";
 import { ServiceWorkerManagerBackground } from "../../components/service-worker-manager";
 import { CookieBanner } from "../../components/cookie-banner";
+import { WEB_APP_URL_ORIGIN } from "../../constants";
 
 const ChakraReactMarkdown = chakra(ReactMarkdown);
-
-const origin =
-  process.env.NEXTAUTH_URL ??
-  (process.env.VERCEL_URL && `https://${process.env.VERCEL_URL}`) ??
-  "https://labelflow.com";
 
 export default function Posts({
   article,
@@ -47,7 +43,7 @@ export default function Posts({
               ]
             : undefined
         }
-        canonical={`${origin}/posts/${article?.slug}`}
+        canonical={`${WEB_APP_URL_ORIGIN}/posts/${article?.slug}`}
         article={{
           authors: article?.author?.name ? [article?.author?.name] : [],
           publishedTime: article?.published_at,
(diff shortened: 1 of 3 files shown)


typescript/web/src/pages/posts/[slug].tsx
```diff
@@ -42,6 +47,12 @@ export default function Posts({
               ]
             : undefined
         }
+        canonical={`${origin}/posts/${article?.slug}`}
+        article={{
+          authors: [article?.author?.name],
```
Comment (bbenoist) [inline typescript/web/src/pages/posts/[slug].tsx:52]: If `article?.author?.name` is `undefined`, you'll have `authors: [undefined]` instead of `authors: []`.
I suggest replacing the value above by either `article?.author?.name ? [article.author.name] : []`

typescript/web/src/pages/posts/[slug].tsx
```diff
@@ -19,6 +19,11 @@ import { CookieBanner } from "../../components/cookie-banner";
 
 const ChakraReactMarkdown = chakra(ReactMarkdown);
 
+const origin =
+  process.env.NEXTAUTH_URL ??
+  (process.env.VERCEL_URL && `https://${process.env.VERCEL_URL}`) ??
+  "https://labelflow.com";
```
Comment (bbenoist) [inline typescript/web/src/pages/posts/[slug].tsx:25]: Can we make this value available globally (e.g. in a `src/constants.ts` file) under the name `ROOT_URL` or similar?

Comment (RaynalHugo) [reply typescript/web/src/pages/posts/[slug].tsx:25]: Done in 4b37dbc44ebcd22d103db4ce6f887ee9e4df7cbc but I named it `WEB_APP_URL_ORIGIN` as it is not specific to articles

typescript/web/src/pages/posts/[slug].tsx
Comment (RaynalHugo) [reply typescript/web/src/pages/posts/[slug].tsx]: Done in 016245f93d83a6f4052af475da30017be328c70c

bbenoist (approved):
LGTM now :)

Issue #630, opened 2021-12-02
Fix dataset import


etiennedupont:
> Ok I'll have a look

In fact I also have problems in local. Some of the images are imported but not all and no labels or classes are created.
![image](https://user-images.githubusercontent.com/16015322/144212528-342b49d2-53b3-40c4-87f6-02bc6b8c89a2.png)

_Originally posted by @nicOwlas in https://github.com/labelflow/labelflow/issues/607#issuecomment-983472663_

etiennedupont:
Fixed in #642

PR #642, opened 2021-12-08
Fix dataset import

etiennedupont:
# Feature

## Resolved issues

#630

### DIFF
--- a/typescript/web/src/pages/posts/[slug].tsx
+++ b/typescript/web/src/pages/posts/[slug].tsx
@@ -36,7 +36,7 @@ export default function Posts({
           article?.image?.url !== null
             ? [
                 {
-                  url: article.image.url,
+                  url: article?.image?.url,
                   alt: "LabelFlow",
                 },
               ]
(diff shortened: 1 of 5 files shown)


Jordanlelay (approved):

typescript/web/src/pages/posts/[slug].tsx
```diff
@@ -36,7 +36,7 @@ export default function Posts({
           article?.image?.url !== null
             ? [
                 {
-                  url: article.image.url,
+                  url: article?.image?.url,
```
Comment (bbenoist) [inline typescript/web/src/pages/posts/[slug].tsx:39]: Won't it be safer to replace `article?.image?.url !== null` by just `article?.image?.url` instead of ignoring `undefined` by using `?` in line 39 (` url: article?.image?.url,`)?

Comment (etiennedupont) [reply typescript/web/src/pages/posts/[slug].tsx:39]: That's correct, let's do that in another PR, I merged right after @Jordanlelay approved

Comment (etiennedupont) [reply typescript/web/src/pages/posts/[slug].tsx:39]: https://github.com/labelflow/labelflow/issues/646

PR #643, opened 2021-12-08
Feature/documentation infrastructure auto polygon

etiennedupont:
# Feature


RaynalHugo:
I think there is a problem with the origin and/or the target of this PR. 
I guess that it was created on `main` and you try to merge it on `feature/workspace`

etiennedupont:
Managed in #624

PR #651, opened 2021-12-10
Add end to end tests in online workspaces

### LINKED ISSUE #553: Create end to end tests for online workspace
- [ ] Port the existing tests for the local workspace to the online workspaces

Blocked by #554
Blocked by #626 
Depends on #519

nacho-carnicero:
# Feature

## Work performed

- added end to end tests for online workspaces
- refactored how end to end tests are organized
- fixed a warning in jest tests where the `clearStore` method would not exist in the apollo client for some test cases

## Resolved issues

Closes #553


cypress/integration/class-selection-popover-online.ts
```diff
@@ -0,0 +1,401 @@
...
+
+  it("right clicks on a label to change its class", () => {
+    // See https://docs.cypress.io/guides/core-concepts/conditional-testing#Welcome-wizard
+    cy.visit(
+      `/cypress-test-workspace/datasets/${datasetSlug}/images/${imageId}?modal-welcome=closed&modal-update-service-worker=update`
+    );
+    cy.get('[aria-label="loading indicator"]').should("not.exist");
+    cy.get('[aria-label="Selection tool"]').click();
+
+    cy.wait(420);
```
bbenoist, line 160: What are you waiting for?
Why `420`?

cypress/integration/class-selection-popover-online.ts
```diff
@@ -0,0 +1,401 @@
...
+      });
+    });
+  });
+
+  it("right clicks on a label to change its class", () => {
+    // See https://docs.cypress.io/guides/core-concepts/conditional-testing#Welcome-wizard
+    cy.visit(
+      `/cypress-test-workspace/datasets/${datasetSlug}/images/${imageId}?modal-welcome=closed&modal-update-service-worker=update`
+    );
+    cy.get('[aria-label="loading indicator"]').should("not.exist");
```
bbenoist, line 157: Wouldn't relying on `id` instead of `aria-label` be a safer way to get component instances? (🤷‍♂️ )

bbenoist (commented):
⚠️  I only did a very quick review (hence the lack of approval).

Overall, I feel like there could be some factorization (e.g. have reusable functions to navigate the website. I mean) but as I said, I didn't read your PR in deep details.

Speaking about reusability, maybe we could find a way to ensure a stronger reference to components by using a shared id value.
i.e. :
```ts
// components-id.ts
// Unique list of all the component IDs in the application
export enum ComponentId {
  loadingIndicator: 'loading-indicator',
  selectionTool: 'selection-tool',
}
```

```tsx
// loading-indicator.tsx
import { ComponentId } from "../component-id.ts";
export const LoadingIndicator: () => (<SomeComp id={ComponentId.loadingIndicator) />);
```

```ts
// class-selection-popover-online.ts
// ...
const get = (id: keyof ComponentId) => cy.get(`[id="${ComponentId[id]}"]`);
// ...
    get("loadingIndicator").should("not.exist");
// ...
```

nacho-carnicero:
@bbenoist thanks for the review, although this is still a WIP it was not ready to review 😅 (you can check on [the board](https://github.com/Labelflow/labelflow/projects/1) which are the ones ready to review). 

I was planning on checking how we could reuse the testing code across the different tests, there's a lot of duplicated code as of now. This was a first test to see if there tests not passing in the CI.

cypress/integration/class-selection-popover-online.ts
nacho-carnicero: The wait commands that wait for `420` are commands that should not be needed but that if removed they introduce some flakiness on the tests. The reason that it's 420 is to be able to identify those wait commands that should not be needed and separate them from other needed wait commands. 

This was not introduced by this PR though, as the purpose of this PR is to take the current existing e2e tests and make them pass on online workspaces

cypress/integration/class-selection-popover-online.ts
nacho-carnicero: The philosophy for why we use `aria-label` is that unless needed we shouldn't add code that is just useful for testing. `aria-label` are useful for accessibility plus they allow us to identify components so that's what we have used so far. I agree that it could become unsafe if we have repeated `aria-labels`, but for now it's not a problem since they are unique in a same view.

Jordanlelay:
I've reviewed it, looks good to me
I'll let @bbenoist give a look as he made some comments earlier

cypress/integration/class-selection-popover.local.ts
```diff
@@ -0,0 +1,156 @@
...
+      name,
+      color,
+      datasetId,
+    },
+  });
+
+  return id;
+};
+
+describe("Class selection popover", () => {
```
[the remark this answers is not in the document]
nacho-carnicero: You're right, I'll do this

typescript/db/src/repository/index.ts
```diff
@@ -134,7 +134,10 @@ export const repository: Repository = {
             await getPrismaClient()
           ).label.update({
             where,
-            data: castObjectNullsToUndefined(label),
+            data: {
+              ...castObjectNullsToUndefined(label),
+              labelClassId: label.labelClassId,
```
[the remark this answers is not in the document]
nacho-carnicero: I had to add this because otherwise we could not set a label to the `None` value in the interface. `None` is not a label class, it just means that the label class is `null`. But since we were casting all the `null` to `undefined` whenever we had a `{labelClass:null}` input to this function it would never be applied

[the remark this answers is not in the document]
bbenoist: OK so this is what I was thinking :) Thanks for the clarification!
So now with your changes, we respect the fact that `labelClassId` is optional in the `Label` model of our Prisma schema. Perfect 👍

cypress/integration/class-selection-popover.local.ts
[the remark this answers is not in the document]
nacho-carnicero: Done in cacb6b9f0cff60164c396d5607ce0248fb143ba9

nacho-carnicero:
@bbenoist I did the refactor to group the mutation declarations for the e2e tests, feel free to have another look when you have time :)

cypress/integration/golden-path.local.ts
```diff
@@ -0,0 +1,5 @@
+import { declareGoldenPathTests } from "./golden-path.common";
+
+describe("Golden path (local)", () => {
+  declareGoldenPathTests({ workspaceSlug: "local", isLocal: true });
```
etiennedupont, line 4: Shouldn't `workspaceSlug =  "local"` be enough to ensure that we are in the `local` case?

nacho-carnicero: True! Done in aa17d50976ca403559dd8d457ad40bc719822d6b

cypress/integration/golden-path.common.ts
```diff
@@ -2,17 +2,15 @@ import imageSampleCollection from "../../typescript/web/src/utils/image-sample-c
 
 export const declareGoldenPathTests = ({
```
bbenoist, line 3: We don't need to use an object anymore here since we've only got a single param 😉

nacho-carnicero: I left it as an object on purpose so that all functions in the graphql definitions have a similar API

bbenoist: OK I see, cool 👍 

That's what I did too in #667 too with the Jest tests 😉 :
* [workspace-name-message.test.tsx](https://github.com/labelflow/labelflow/blob/feature/delete-workspace/typescript/web/src/components/workspace-name-input/workspace-name-message.test.tsx#L83)
* [create-workspace-modal.test.tsx](https://github.com/labelflow/labelflow/blob/feature/delete-workspace/typescript/web/src/components/workspace-switcher/workspace-creation-modal/create-workspace-modal.test.tsx#L95)

I even used the same names for `declareTests`, the `TestCase` type and the `TEST_CASES` constant so that files can be copy-pasted with more ease when adding a new unit-test 🚀  Maybe we could do the same here too?

nacho-carnicero: I did something similar in 7615c26d8f0457a529d44ec4925c1cc9a5b60507, let me know if it's what you were expecting!

bbenoist: Yes that's perfect 👌 Thank you 🙂

bbenoist (approved):
Good job 😎

PR #652, opened 2021-12-10
Re-enable jest tests

RaynalHugo:
# Bug Fix

## Work performed

- Fix jest config. 2 Glob patterns were incorrect, preventing running tests in `common-resolvers` and `web`.
- Due to this change, we had unnoticed failing tests. I fixed those tests.

## Results

All tests are now launched again. And they are all passing.

## Problems encountered

Some tests aren't well written. They test side effects, making them really prone to fail if we change the implementation, but not the function signatures. I didn't refactor those tests as I wanted this PR to be merged ASAP. But it would be nice to take the time to rewrite them someday.


typescript/web/src/components/labeling-tool/openlayers-map/__tests__/draw-bounding-box-and-polygon-interaction.tsx
```diff
@@ -132,9 +132,14 @@ it("is possible to undo the creation of the label", async () => {
 
   await useUndoStore.getState().undo();
 
+  // @ts-ignore
```
bbenoist, line 135: @RaynalHugo Why is there a `ts-ignore` here? Can't you use TS typings to fix the compiler issue?

RaynalHugo: Typescript doesn't understand itself that we called jest.mock on the apollo client. 

I removed the ts-ignore, but replaced it with `as jest.Mock` on the apollo client. Apparently, that's the standard way to deal with mocks. 

It's not really type-safe. After checking what it does, everything after `calls[0][0]` is typed as any. 
I think `jest.Mock` is a generic and I could somehow type this. But I don't think this is necessary to go further on this. It's only used on tests and that is basically what we want to assert on the `expect` on the following lines.

RaynalHugo: @bbenoist I am merging the PR now as it is blocking, but, if you want, we can continue discussing this here and implement the needed changes on another PR.

PR #657, opened 2021-12-15
Deploy schema in CI

### LINKED ISSUE #584: Deploy the schema in the CI
## Objectives
Allow to deploy automatically new schema

## Tasks

- [ ] CI need another port to connect directly to postgres (and bypass pgbouncer): https://community.redwoodjs.com/t/guide-migrations-connection-pool-urls-supabase/2505
- [ ] Follow prisma documentation for the good practice on how to deploy it in the CI (@RaynalHugo for the link to the doc)

Blocked by #621

nacho-carnicero:
# Feature

## Work performed

- added scripts to migrate prisma to DB only in production deployments
- updated Vercel build command to call `build:vercel`

## Caveats

I did not manage to call directly the shell script to conditionally execute the migration from a command in the `package.json` file. I would get the following error:

```
Unbound variable "MIGRATE_DATABASE"
Unbound variable "MIGRATE_DATABASE"
command not found: else
```

As a workaround I put the script in a bash file that is called from the Vercel build command. This is the best solution that I could find, if you have other suggestions please let me know!

## Results

In order to test this I set `MIGRATE_DATABASE="this-is-not-a-db-url"` for preview deployments, and as expected the build script breaks during the migration part (see the relevant logs [here](https://vercel.com/labelflow/labelflow/FD59C6Xxi6znyat6VPoedqMsmQRi)). I expect therefore this to work with the production deployment since for those the `MIGRATE_DATABASE` env var is correctly set.

## Resolved issues

Closes #584


typescript/db/package.json
```diff
@@ -10,7 +10,8 @@
     "stop": "docker compose down --volumes --remove-orphans",
     "deploy": "prisma migrate deploy",
     "dev:push": "prisma db push",
-    "generate": "prisma generate"
+    "generate": "prisma generate",
+    "resolve": "prisma migrate resolve --applied  20210601170000_add_strapi"
```
RaynalHugo, line 14: Do we want to keep this?

typescript/db/deploy-vercel.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+if [[ -n $MIGRATE_DATABASE ]]; 
+ then POSTGRES_EXTERNAL_URL=$MIGRATE_DATABASE yarn deploy && POSTGRES_EXTERNAL_URL=$POSTGRES_EXTERNAL_URL; 
+ else echo 'Skipping migration. No $MIGRATE_DATABASE' && echo $MIGRATE_DATABASE;
```
RaynalHugo, line 4: Do we want to keep the second `echo`?

typescript/db/package.json
nacho-carnicero: Wow not at all, good catch, will remove it!

typescript/db/deploy-vercel.sh
nacho-carnicero: I left it so that we can more easily track if a migration is happening or not, but if you feel it's not used I can remove it

RaynalHugo: No, that's ok. I wanted to know if this was only to debug it when you implemented this. But you're right, it may be useful later on.

typescript/db/package.json
nacho-carnicero: Done in c93498eb53844378be5e7e84c6d78c34cd5699a5

nacho-carnicero:
@RaynalHugo let me know if you think this is mergeable now! :)

RaynalHugo (approved):

Issue #603, opened 2021-11-22
Create 1 connection to the DB per upload instead of per file


nicOwlas:
**Context**
- When doing an upload, there is 1 connection to the DB per file uploaded which saturates PGBouncer

**Objective**
- Robustify how many uploads can happen in parallel without saturating PGBouncer

**Solution 1**
- [ ] Call `getUploadTarget` 1 time per file
  - [ ] If `getUploadTarget` calls the DB, then code and call a new resolver getUploadTargets that returns many upload URLs
  - [ ] If no keep as is
- [ ] Code and call a new resolver `createManyImages` with all the urls returned by `getUploadTarget` (thumbnails are created in the `createImage` mutation). Upload directly to AWS thanks to the presigned URL returned by AWS through the grapql resolvers (hosted on Vercel)
- [ ] Eventually batch getUploadTarget if not enough

Blocked by #607 
Blocked by #626

RaynalHugo:
Blocked by #626 
I can't upload images on the distant workspace, making this hard to test

RaynalHugo:
Done in #603

Issue #366, opened 2021-08-20
Simplify the code to keep data consistent in Apollo


frinyvonnick:
> Something is wrong with the refetch of projects ^^
> 
>  Screen.Recording.2021-08-19.at.12.23.04.mov

I fixed it in 63efe7e but I'm not really happy with my solution. The cache management is not really straightforward. I would like your thoughts on this @crubier.

_Originally posted by @frinyvonnick in https://github.com/Labelflow/labelflow/issues/356#issuecomment-902534227_

PR #662, opened 2021-12-16
Feature/batch image creation

RaynalHugo:
# Feature

## Work performed

Batch images creation to avoid taking too many connections to the database.

## Results

This PR does:
- Add a new mutation `createManyImages`
- Refactor some logic on the image creation resolvers to make it easier to read and reuse
- Implement the new `createManyImages` resolvers on the upload modal for both URL and file uploads
- Refactor the upload code on the modals to make it easier to read and reuse

Also, it:
- Updates some `tsconfig.json` files to add strict mode and `strictNullCheck`.
- Removes a security breach where we were sending the users cookies to potential external hosts.

## Problems encountered

- Code was messy and not easy to update. Had to refactor a lot of things.
- There was and still is some duplication of the logic in some places.
- COCO export wasn't working locally
- Thumbnail generation was updating images before creating them 🤔 

## Caveats

- This PR reduces the number of connections done to the database by further reducing the concurrency of mutations in parallel. This is ok as now each mutation create 10 images. But it could maybe be faster if we increased this concurrency. More details on the compromise and how to adjust it here: `typescript/web/src/components/import-button/import-images-modal/constants.ts`
- The progress is updated per batch. From the user perspective, it looks like that it alternates between being stuck and being fast. It would be better to handle this differently.
- Thumbnail processing is done "synchronously" during the `createManyImages` mutation, making it slow at scale. #679 
- Error handling is done per batch, meaning that if one image of the batch fails, the entire batch fails. #688 
- I didn't remove the `UploadTargetDirect` way to upload, but it's not used anymore, it would be nice to properly deprecate it. (Not done in this issue as it would take quite some time to do it well)
- The create image mutations accepts thumbnail urls as inputs. Contrary to the `externalUrl` way to import, the external thumbnail urls are NOT transferred to our storage. I'll create an issue about that.

## Resolved issues

#603 

## Newly raised issues

- #679
- #687 
- #688 
- @nicOwlas, didn't you create an issue about handling image upload progress? I can't find it, else I'll create it. 
- Creating one for uploadTargetDirect deprecation
- Creating an issue about storing external thumbnails into our storage


typescript/db/src/repository/index.ts
```diff
@@ -37,6 +37,17 @@ export const repository: Repository = {
...
+    addMany: async ({ images, datasetId }, user) => {
+      await checkUserAccessToDataset({ where: { id: datasetId }, user });
+
+      const prisma = await getPrismaClient();
+
+      await prisma.image.createMany({
+        data: images,
+      });
+
+      return images.map((image) => image.id);
```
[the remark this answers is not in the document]
bbenoist: From what I've seen in the resolver, the `id` property is not created by the caller.
And not not required in the GraphQL schema.
Where does its value comes from?

data/__generated__/schema.graphql
```diff
@@ -164,6 +164,30 @@ input ImageCreateInput {
   metadata: JSON
 }
 
+input ImageCreateManyInput {
+  images: [ImageCreateManySingleImage!]!
+  datasetId: ID!
+}
+
+input ImageCreateManySingleImage {
+  id: ID
```
bbenoist, line 173: Why would we allow to force an ID property?

data/__generated__/schema.graphql
```diff
@@ -164,6 +164,30 @@ input ImageCreateInput {
...
+  createdAt: DateTime
```
bbenoist, line 174: This should be set on server-side.

typescript/web/src/components/import-button/import-images-modal/__tests__/import-images-modal.tsx
```diff
@@ -109,7 +109,9 @@ test("should display the number of valid images", async () => {
   ).not.toBeInTheDocument();
 });
 
-test("should update completed number as valid images are uploaded", async () => {
+// Currently, the progress is done per batch. Leaving this test on purpose but skipped
+// as this is likely something we want to enable again.
+test.skip("should update completed number as valid images are uploaded", async () => {
   await renderModalAndImport();
```
bbenoist, line 115: If it's dead code, just remove it 🔥

typescript/web/src/components/import-button/import-images-modal/constants.ts
```diff
@@ -0,0 +1,23 @@
...
+ * Maximum number of batches to perform in parallel.
+ * Increasing this number should reduce potentials bottlenecks,
+ * and thus, make the upload faster.
+ * For example, if the user uploads faster than the images are processed on
+ * the server, then the upload would be delayed until one batch is fully finished.
+ *
+ * However, increasing this number will also increase the number
+ * of connections to the database.
+ */
+export const CONCURRENCY = 2;
```
bbenoist, line 23: 👍  For this whole file

typescript/web/src/components/import-button/import-images-modal/modal-dropzone/import-dropped-files/import-datasets.tsx
```diff
@@ -0,0 +1,95 @@
...
+  file,
+}: {
+  apolloClient: ApolloClient<any>;
+  datasetId: string;
+  workspaceId: string;
+  file: Blob;
+}) => {
+  const key = `${workspaceId}/${datasetId}/import-datasets/${uuidv4()}-${new Date().toISOString()}.${mime.extension(
+    file.type
+  )}`;
```
bbenoist, line 35: What's the purpose of this key?
Can you make is simpler by using intermediary variables `id`, `now` and `extension`?

typescript/web/src/components/import-button/import-images-modal/modal-dropzone/import-dropped-files/import-datasets.tsx
```diff
@@ -0,0 +1,95 @@
...
+    }
+  }
+`;
+const importDataset = async ({
+  apolloClient,
+  datasetId,
+  workspaceId,
+  file,
+}: {
+  apolloClient: ApolloClient<any>;
```
bbenoist, line 28: Can't we give ApolloClient the correct type to infer? (or use the `useMutation` hook)

typescript/web/src/components/import-button/import-images-modal/modal-dropzone/import-dropped-files/import-datasets.tsx
```diff
@@ -0,0 +1,95 @@
...
+          },
+        },
+      },
+    },
+    refetchQueries: [
+      "getDatasetData",
+      "getImageLabels",
+      "getLabel",
+      "countLabelsOfDataset",
+      "getDatasetLabelClasses",
```
bbenoist, line 57: While it's necessary to tell which queries are invalidated, I'm afraid we'll have a lot of bugs when forgetting to add new queries in such arrays. Isn't there a way to make it more future-proof?

typescript/web/src/components/import-button/import-images-modal/modal-dropzone/import-dropped-files/import-datasets.tsx
```diff
@@ -0,0 +1,95 @@
...
+  apolloClient,
+  setFileUploadStatuses,
+}: {
+  datasets: DroppedFile[];
+  datasetId: string;
+  workspaceId: string;
+  apolloClient: ApolloClient<any>;
+  setFileUploadStatuses: SetUploadStatuses;
+}) => {
+  return await Bluebird.Promise.map(
```
bbenoist, line 78: Why do you use Bluebird?

typescript/web/src/components/import-button/import-images-modal/modal-dropzone/import-dropped-files/import-datasets.tsx
```diff
@@ -0,0 +1,95 @@
...
+  mutation importDataset(
+    $where: DatasetWhereUniqueInput!
+    $data: DatasetImportInput!
+  ) {
+    importDataset(where: $where, data: $data) {
+      error
+    }
+  }
+`;
+const importDataset = async ({
```
bbenoist, line 22:
```suggestion
`;

const importDataset = async ({
```

typescript/web/src/components/import-button/import-images-modal/modal-dropzone/import-dropped-files/import-images.ts
```diff
@@ -0,0 +1,109 @@
...
+  workspaceId: string;
+  apolloClient: ApolloClient<any>;
+}) => {
+  return await Promise.all(
+    batch.map(async ({ file }, fileIndex) => {
+      const createdAtDate = new Date();
+      createdAtDate.setTime(
+        firstUploadDate.getTime() + batchIndex * BATCH_SIZE + fileIndex
+      );
+      const createdAt = createdAtDate.toISOString();
```
bbenoist, line 42: Why to we need this?

typescript/web/src/components/import-button/import-images-modal/modal-dropzone/import-dropped-files/import-images.ts
```diff
@@ -0,0 +1,109 @@
...
+      const key = `${workspaceId}/${datasetId}/${uuidv4()}.${mime.extension(
+        file.type
+      )}`;
```
bbenoist, line 45: What's the purpose of this key? Why is it different from the one in `import-datasets.tsx`

typescript/web/src/components/import-button/import-images-modal/modal-dropzone/import-dropped-files/index.tsx
```diff
@@ -0,0 +1,48 @@
...
+  datasetId: string;
+  setFileUploadStatuses: SetUploadStatuses;
+  apolloClient: ApolloClient<any>;
+}) => {
+  const acceptedFiles = files.filter((file) => isEmpty(file.errors));
+
+  const [images, datasets] = partition(
+    // Accepted files are either `images/*`, `application/json` or `application/zip`
+    // The two last are treated as datasets
+    (acceptedFile) => acceptedFile.file.type.startsWith("image"),
```
bbenoist, line 29:
```suggestion
    (acceptedFile) => acceptedFile.file.type.startsWith("image/"),
```

typescript/web/src/components/import-button/import-images-modal/modal-dropzone/import-dropped-files/index.tsx
```diff
@@ -0,0 +1,48 @@
...
+  files: DroppedFile[];
+  workspaceId: string;
+  datasetId: string;
+  setFileUploadStatuses: SetUploadStatuses;
+  apolloClient: ApolloClient<any>;
+}) => {
+  const acceptedFiles = files.filter((file) => isEmpty(file.errors));
+
+  const [images, datasets] = partition(
+    // Accepted files are either `images/*`, `application/json` or `application/zip`
```
bbenoist, line 27:
```suggestion
    // Accepted files are either `image/*`, `application/json` or `application/zip`
```

typescript/web/src/components/import-button/import-images-modal/modal-dropzone/modal-dropzone.tsx
```diff
@@ -153,171 +67,35 @@ export const ImportImagesModalDropzone = ({
...
+      setFileUploadStatuses,
+      apolloClient,
+      onUploadStart,
+      onUploadEnd,
+    ]
+  );
+
+  useEffect(() => {
+    if (isEmpty(files)) return;
+    if (!datasetId) return;
```
bbenoist, line 96:
```suggestion
    if (isEmpty(files)) || !datasetId) return;
```

typescript/web/src/components/import-button/import-images-modal/types.ts
```diff
@@ -22,3 +22,13 @@ export type DroppedUrl = {
  * the file being uploaded
  */
 export type UploadStatuses = Record<string, boolean | string>;
+
+/**
+ * Generic type of an useState setter function
+ */
+type SetState<T> = (newStateOrUpdater: T | ((oldState: T) => T)) => void;
```
bbenoist, line 29: Isn't there a standard type already covering this case?

typescript/web/src/connectors/repository/image-processing.ts
```diff
@@ -1,3 +1,7 @@
+// It seems that this file is almost a copy past of `typescript/db/src/repository/image-processing.ts`
+// We should either: merge/factorize the two files, or let them diverge and have a dedicated
+// version for each repository.
+
```
bbenoist, line 4: 👍  for the merge

typescript/web/src/connectors/repository/image-processing.ts
```diff
@@ -10,6 +14,8 @@ const maxImageSizePixel: { [mimetype: string]: number } = {
   "image/png": 60e6,
 };
 
+type ThumbnailSizes = 20 | 50 | 100 | 200 | 500;
+
```
bbenoist, line 18: This type is already declared in another module

typescript/web/src/connectors/repository/utils/list.ts
```diff
@@ -31,6 +31,12 @@ export const list =
   ): Promise<Entity[]> => {
     const table = await getTable();
     if (where) {
+      if ("id" in where && "in" in where.id) {
+        return (await table.bulkGet(where.id.in)).filter(
+          (entity) => entity !== undefined
+        ) as Entity[];
+      }
+
```
bbenoist, line 39: Which specific case does this covers?

typescript/web/src/utils/upload-file.ts
```diff
@@ -0,0 +1,74 @@
...
+        downloadUrl
+      }
+    }
+  }
+`;
+
+/**
+ * Retrieves the upload url and download url for the given key.
+ * @param key key to identify the file in the storage
+ * @param apolloClient The apollo client to use for the mutation
```
bbenoist, line 22:
```suggestion
 * @param key - Unique key to identify the file in the storage
 * @param apolloClient - The apollo client to use for the mutation
```

typescript/web/src/utils/upload-file.ts
```diff
@@ -0,0 +1,74 @@
...
+};
+
+/**
+ * Uploads a file to the storage at the specify key.
+ * Gets the upload url from the server.
+ *
+ * @param file the blob or file to upload
+ * @param key key to identify the file in the storage
+ * @param apolloClient The apollo client to use for the mutation
+ * @returns the url to download the file
```
bbenoist, line 52:
```suggestion
 * Uploads a file to the storage at the specify key
 * @remarks
 * Gets the upload url from the server.
 *
 * @param file - The blob or file to upload
 * @param key - Unique key to identify the file in the storage
 * @param apolloClient - The apollo client to use for the mutation
 * @returns The url to download the file
```

bbenoist (changes requested):

typescript/web/src/components/import-button/import-images-modal/modal-dropzone/import-dropped-files/import-datasets.tsx
RaynalHugo: That's the standard library we used until now when we want to map on an array with a limited concurrency.

typescript/web/src/components/import-button/import-images-modal/modal-dropzone/import-dropped-files/import-images.ts
RaynalHugo: The image order is based on their created at. 
To keep the order given during the upload, we artificially add 1ms between 2 images. (That how it was done so far, it's not a behaviour I added in this PR)

typescript/web/src/components/import-button/import-images-modal/modal-dropzone/import-dropped-files/import-images.ts
RaynalHugo: To put and retrieve the image in the storage.

It's different that the one in import dataset because datasets imports need to upload a zip of json file to the storage, then this file is processed on the server side. 

As it's 2 different behaviours, I found it useful to add a difference between those 2 processes.

typescript/web/src/components/import-button/import-images-modal/types.ts
RaynalHugo: React already provides but it's something like `React.Dispatch<React.SetStateAction<T>>`. I felt like it's too tied to `useStates`, but I can use this one if you prefer.

typescript/web/src/connectors/repository/image-processing.ts
RaynalHugo: Done in 5be069f91fcdabd84671f60ec3842daa55b883da

typescript/web/src/connectors/repository/utils/list.ts
RaynalHugo: It's used here: https://github.com/labelflow/labelflow/blob/5be069f91fcdabd84671f60ec3842daa55b883da/typescript/common-resolvers/src/image/index.ts#L150

typescript/common-resolvers/src/image/import-and-process-image.ts
```diff
@@ -0,0 +1,264 @@
...
+  // eslint-disable-next-line no-underscore-dangle
+  if (uploadTarget.__typename !== "UploadTargetHttp") {
+    throw new Error(
+      "This Server does not support file upload. You can create images by providing a `file` directly in the `createImage` mutation"
+    );
+  }
+
+  await repository.upload.put(uploadTarget.uploadUrl, blob, req);
+
+  return uploadTarget.downloadUrl;
```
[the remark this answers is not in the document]
RaynalHugo: Deduplicated in 9124f83f510692b0a598bfb6a2475da42fa09256

typescript/web/src/components/import-button/import-images-modal/modal-dropzone/import-dropped-files/import-datasets.tsx
RaynalHugo: Done in 3438be54a3dff8dd5d3955943810b3d7416796b7

typescript/web/src/components/import-button/import-images-modal/modal-dropzone/import-dropped-files/import-datasets.tsx
RaynalHugo: The purpose of this key is to upload the file to import (zip or json) into the storage. Then we can process it server side.

RaynalHugo: I did the change on the syntax in 22e1151b6f3af9e0d8da1a51ce3784657e231242

typescript/web/src/components/import-button/import-images-modal/modal-dropzone/import-dropped-files/import-datasets.tsx
RaynalHugo: That's not something I introduced in this PR. 
Changing this would be quite difficult. It implies to rely only on cache updates. 

At some point we discussed migrating from refetch queries to cache updates (in https://github.com/labelflow/labelflow/issues/366).

RaynalHugo (commented):
@bbenoist 
I applied all the changes you suggested. Thanks for the thorough review. 

I left some conversations opened where I felt that it would be useful to have your answer/validation :)

data/__generated__/schema.graphql
RaynalHugo: It's sometimes useful to generate id on the client side. For example, with optimistic responses, or for the undo/redo

data/__generated__/schema.graphql
RaynalHugo: See other comment, that's what we use to keep images in the same order as when they were uploaded.

data/image.graphql
```diff
@@ -19,6 +19,31 @@ input ImageCreateInput {
   metadata: JSON
 }
 
+# Same as ImageCreateInput but without datasetId
```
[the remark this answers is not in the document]
bbenoist: Sorry, my comment was the very-first of my code-review and I misunderstood the purpose of this type. I should have reviewed my code-review 😅

typescript/web/src/components/import-button/import-images-modal/types.ts
bbenoist: Yes, it is equivalent to `Dispatch<SetStateAction>`:

```ts
    // Unlike the class component setState, the updates are not allowed to be partial
    type SetStateAction<S> = S | ((prevState: S) => S);
    // this technically does accept a second argument, but it's already under a deprecation warning
    // and it's not even released so probably better to not define it.
    type Dispatch<A> = (value: A) => void;
```

It's still preferable to reuse the existing types instead of redefining them. Even more in such a specific file.

 Given that there is no shorthand for `Dispatch<SetStateAction>`, you could still add (in e.g. `typescript/web/src/utils/types/set-state.ts` - with a proper `index.ts` chain so that we don't have to use the full path to import it 😉 ) the following type definition if you believe it'll be too much repeated in the code:

 ```ts
 import { Dispatch, SetStateAction } from "react";
 
 type SetState<TValue> = Dispatch<SetStateAction<TValue>>;
 ```

typescript/web/src/components/labeling-tool/openlayers-map/classification/classification-tag.tsx
```diff
@@ -28,7 +28,7 @@ export const ClassificationTag = ({
   labelClasses: LabelClassItem[];
   selectedLabelId: string | null;
   setSelectedLabelId: (labelId: string | null) => void;
-  client: ApolloClient<any>;
+  client: ApolloClient<object>;
```
bbenoist, line 31: While it's cool to remove `any`, don't we really have something more strong?

typescript/web/src/connectors/repository/image-processing.ts
```diff
@@ -96,82 +92,39 @@ export const processImage: Repository["imageProcessing"]["processImage"] =
...
+        mimetype: mimetype ?? image.getMIME(),
       }),
+      thumbnail20Url: thumbnail20Url ?? (await generateThumbnailFromSize(20)),
+      thumbnail50Url: thumbnail50Url ?? (await generateThumbnailFromSize(50)),
+      thumbnail100Url:
+        thumbnail100Url ?? (await generateThumbnailFromSize(100)),
+      thumbnail200Url:
+        thumbnail200Url ?? (await generateThumbnailFromSize(200)),
+      thumbnail500Url:
+        thumbnail500Url ?? (await generateThumbnailFromSize(500)),
```
bbenoist, line 128: From what I understand, thumbnails will be generated sequentially here. Isn't it?

typescript/web/src/connectors/repository/utils/list.ts
bbenoist: That's not exactly what I meant by which specific case 😅 
If the `in` keyword is a feature that Dexie is missing over Prisma, can we fix the problem more globally and support it closer to Dexie?

typescript/web/src/utils/upload-file.ts
```diff
@@ -0,0 +1,75 @@
...
+ */
+export const uploadFile = async ({
+  file,
+  key,
+  apolloClient,
+}: {
+  file: Blob;
+  key: string;
+  apolloClient: ApolloClient<object>;
+}) => {
```
bbenoist, line 63:
```suggestion
export const uploadFile = async (
  file: Blob,
  key: string,
  apolloClient: ApolloClient<object>
) => {
```

typescript/web/src/connectors/repository/index.ts
```diff
@@ -42,6 +42,17 @@ export const repository: Repository = {
...
+    addMany: async ({ images, datasetId }) => {
+      const imagesToAdd = images.map((image) =>
+        addIdIfNil({ ...image, datasetId })
+      );
+
+      return await (
+        await getDatabase()
+      ).image.bulkAdd(imagesToAdd, {
+        allKeys: true,
+      });
```
bbenoist, line 54:
```suggestion
      const newImages = images.map((image) =>
        addIdIfNil({ ...image, datasetId })
      );
      const db = await getDatabase();
      return await db.image.bulkAdd(newImages, { allKeys: true, });
```

bbenoist (approved):
My new comments are only regarding good practices and not critical. Feel free to merge this PR directly if you want.

typescript/web/src/components/labeling-tool/openlayers-map/classification/classification-tag.tsx
RaynalHugo: Ah sorry, I justified this change, but on the commit message, so it doesn't appear here.

First I went with `ApolloClient<NormalizedCacheObject>` which is the exact type of our clients.
However, it was conflicting with `useApolloClient` which returns an `ApolloClient<object>`.

So, I guessed it was safe to use this like this (this parameter only define the shaped of the cache).

nicOwlas:
Just tested to upload 2000 items on this release. It's much much faster. 👏

nicOwlas:
I ran into the same disconnection issue #681 when trying to upload a 2nd batch.

typescript/web/src/connectors/repository/image-processing.ts
RaynalHugo: Good catch, I change it here 0b26d6e972ecce8922c4940abe4fb3a8b5cacbf3

PR #653, opened 2021-12-12
Loading State on Dataset Page

2Clutch:
# Feature

We would like to improve the user experience of the dataset page by replacing the current `Create new dataset` card with a skeleton card.

## Resolved issues

Fixes #627

Issue #627, opened 2021-12-01
Display a loading state on the dataset page


nicOwlas:
## Tasks
- Implement a spinner in place of the 1st dataset card to display datasets are loading ([chakra component here](https://chakra-ui.com/docs/feedback/spinner))

## Mockup

![image](https://user-images.githubusercontent.com/16015322/145822801-4c22f34e-2552-4dc8-9b79-a9839ad42d0c.png)

```
<Spinner
  thickness='4px'
  speed='0.65s'
  emptyColor='gray.200'
  color='#31CECA'
  size='xl'
/>
```

![image](https://user-images.githubusercontent.com/16015322/145822752-aee43c43-25ee-47c1-9845-b5b176051627.png)

2Clutch:
Hello. Found your project on `goodfirstissues.com`, and I'd like to take on this task if no one else is working on it.

nicOwlas:
@2Clutch it's yours! Thanks a lot! I updated the description to have a skeleton instead of a spinner.
Once you start working on it, could you assign yourself move it in the `In progress` column in the [project board here](https://github.com/labelflow/labelflow/projects/1)?

 Let us know if you have any questions.

2Clutch:
Thank you @nicOwlas. I sincerely appreciate it.

If I understand properly, the workflow should be:
1. User click on `Create a new Project`
2. User selects an image to upload
3. Display `skeleton` as upload is being processed.
4. Display image once upload has been processed successfully.

Am I accurate, in my description of what I believe is needed to address this issue?

nicOwlas:
@2Clutch the workflow is much simpler for this issue:
1. The user opens the dataset gallery url (today: https://labelflow.ai/local/datasets)
2. The dataset cards are being loaded and in the meantime, we display to the user the dataset card skeleton
3. Once the actual dataset card is loaded we replace the skeleton with its dataset card

Today, as there is no skeleton, if the user's connection is slow, he/she would otherwise only see the `Create a new dataset` card and nothing else. He/she may think his/her datasets were lost where in fact they are being loaded in the background. Displaying the dataset card skeleton allows showing to the user that some elements on the page are being loaded.

You can open a PR and commit regularly so that we give you feedbacks along the way. When do you think you'll have time to start the issue?

2Clutch:
I'll get started over the weekend. I cloned the project and started going through the initial setup.

2Clutch:
[...]

PR #668, opened 2021-12-17
Add spinner on loading datasets and add reusable card box component

Jordanlelay:
# Feature

## Work performed

- Add a spinner when the datasets are loading
- Add a reusable component wrapping the `NewDatasetCard`, the `DatasetCard` and the `LoadingCard` components
- Fix key error in tab bar

## Results

A spinner will be displayed when the datasets are loading
We can now use the `DatasetCardBox` component to obtain a flexible `Box` having the same size as the other "Cards". Modifying the `DatasetCardBox` component will update all 3 components quoted before

## Resolved issues

This PR will solve #627
This PR will close #653


nacho-carnicero (approved):
LGTM :)

Issue #666, opened 2021-12-17
Several classes can have the same color


Jordanlelay:
**Describe the bug**
When creating a new label class, it is possible to generate several times the same label color since we can reorder them

**To Reproduce**
1. Go to the labeling tool and create a class, you will have to classes: Horse and the one you just generated
1. Go the classes page, Horse class should be first
1. Reorder the 2 classes and put Horse at last
1. Go back to the labeling tool and create an other class
1. The error is that the 2 classes generated have the same color

**Expected behavior**
Each class you generate (until you reach the limit of the hardcoded classes color) should have a unique color

**Screenshots**
![Capture d’écran 2021-12-17 à 11 25 27](https://user-images.githubusercontent.com/43350109/146530232-4df15a3a-42b7-434c-be2b-6cfea26be194.png)

**Desktop (please complete the following information):**
All versions affected

**Smartphone (please complete the following information):**
All versions affected

Jordanlelay:
We could potentially ignore this bug if we let the users select the color themselves. Currently they cannot change the color once the class created

nicOwlas:
Good catch. We can also fix the logic how the colour is assigned to a class. There is a list of predefined colours and we should avoid picking an already assigned colour.

vincent-lecrubier-skydio:
😈

Jordanlelay:
Closed by #673

PR #673, opened 2021-12-20
Update label class color generator

Jordanlelay:
# Feature

## Work performed

- Update the way label class colors are generated to ensure we don't generate the same color two times in a row
- Add a utils package for shared functions across packages and avoid duplication

## Results

We can now generate label classes, reorder, delete them and generate again without having colors duplication

## Caveats

You will have to run a `yarn install` in order to have a link established with the newly created package

## Resolved issues

This PR solves #666


typescript/utils/class-color-generator.ts
```diff
@@ -0,0 +1,110 @@
...
+  });
+
+  return hashMap;
+};
+
+const getColorNotAttributed = (
+  numberOfAttributedColors: number,
+  attibutedColors: string[]
+) => {
+  for (let i = 0; i < numberOfAttributedColors; i += 1) {
```
valerian, line 58: This wrongly assumes that `attributedColors` contains no duplicate color, or that they are in an order matching the order of `hexColorSequence`.
In this situation, `numberOfAttributedColors` should be replaced with `hexColorSequence.length`, with some other minor tweaks.

typescript/utils/class-color-generator.ts
```diff
@@ -0,0 +1,110 @@
...
+    if (!attibutedColors.includes(hexColorSequence[i])) {
+      return hexColorSequence[i];
+    }
+  }
+
+  return hexColorSequence[numberOfAttributedColors];
```
valerian, line 64: It's dangerous to assume `numberOfAttributedColors` will always be an index of `hexColorSequence`!

typescript/utils/class-color-generator.ts
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  if (numberOfAttributedColors < hexColorSequence.length) {
+    // If number of attributed colors is strictly less than the number of available colors, we
+    // want the first color not attributed already
+    return getColorNotAttributed(numberOfAttributedColors, attibutedColors);
+  }
+
+  // Last case is all colors have been attributed at least once, but not a number of times that
+  // is a multiple of hexColorSequence. We want the first color with the smallest number of attributions
```
valerian, line 104: Again, you don't know at this point if all colors have been attributed at least once, because `attibutedColors` might have duplicates, which are unaccounted for in the previous conditions.

typescript/utils/class-color-generator.ts
```diff
@@ -0,0 +1,110 @@
...
+
+const createColorHashMap = (attibutedColors: string[]) => {
+  const hashMap: { [key: string]: number } = {};
+  attibutedColors.forEach((attributedColor) => {
+    if (attributedColor in hashMap) {
+      hashMap[attributedColor] += 1;
+    } else {
+      hashMap[attributedColor] = 1;
+    }
+  });
```
valerian, line 49: According to my other comments, you are not sure here that `attibutedColors` will contain all the possible colors from `hexColorSequence`, and it will cause problems later.
You should first initialize the hashMap to zero for all `hexColorSequence` values.

typescript/utils/class-color-generator.ts
```diff
@@ -0,0 +1,110 @@
...
+  const hashMap: { [key: string]: number } =
+    createColorHashMap(attibutedColors);
+
+  const targetNumberOfOccurrences = Math.floor(
+    numberOfAttributedColors / hexColorSequence.length
+  );
+
+  for (let i = 0; i < overflow; i += 1) {
+    if (hashMap[hexColorSequence[i]] === targetNumberOfOccurrences) {
+      return hexColorSequence[i];
```
valerian, line 81: According to my other comments, both `overflow` and `hashMap` are unreliable in this situation.
Also why use a calculated `targetNumberOfOccurrences` instead of just keeping the first found lowest number in the `hashMap`?

valerian (changes requested):
About the `getNextClassColor`, there are several issues I pointed out in my code comments which make it prone to returning wrong colors in some cases.

In the end, I think that you could simplify it and use only `getFirstColorLessAttributed` and build a hashMap that is initialized to zero for all `hexColorSequence` values.
You won't need `numberOfAttributedColors`, `overflow`, and `getColorNotAttributed` anymore.

I see that you tried to optimize performance by avoiding to use the `hashMap` when not necessary, but there are too many special cases to handle, and the zero-initialized `hashMap` solution handles all of them, without really using that much resources in the end.

typescript/utils/class-color-generator.ts
```diff
@@ -0,0 +1,110 @@
+export const hexColorSequence = [
```
bbenoist, line 1:
```suggestion
export const LABEL_CLASS_COLOR_PALETTE = [
```

typescript/utils/class-color-generator.ts
```diff
@@ -0,0 +1,110 @@
...
+  "#E11D48",
+  "#D97706",
+  "#16A34A",
+  "#0891B2",
+  "#4F46E5",
+  "#C026D3",
+];
+
+// It's charkra's gray.200
+export const noneClassColor = "#E2E8F0";
```
bbenoist, line 39: Do we need to define a color for `none` at such a low-level?

typescript/utils/class-color-generator.ts
bbenoist:
```suggestion
const countUniqueColors = (colors: string[]) => {
  return colors.reduce<Record<string, number>>(
    (counts, color) => ({
      ...counts,
      [color]: (counts[color] ?? 0) + 1,
    }),
    {}
  );
};
```

typescript/utils/class-color-generator.ts
```diff
@@ -0,0 +1,110 @@
...
+  return hexColorSequence[overflow];
+};
+export const getNextClassColor = (attibutedColors: string[]): string => {
+  const numberOfAttributedColors = attibutedColors.length;
+  const overflow = numberOfAttributedColors % hexColorSequence.length;
+
+  if (overflow === 0) {
+    // If the rest of the division is 0, all colors have been attributed once or several
+    // times, so we return the first color of the array
+    return hexColorSequence[0];
```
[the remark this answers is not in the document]
Jordanlelay: Yes, that's right

typescript/utils/class-color-generator.ts
Jordanlelay: I didn't assume it, this method is called only when `numberOfAttributedColors` is strictly less than `hexColorSequence.length` 🙂 
Anyway, it will be removed as discussed

typescript/utils/class-color-generator.ts
```diff
@@ -0,0 +1,110 @@
...
+
+  for (let i = 0; i < overflow; i += 1) {
+    if (hashMap[hexColorSequence[i]] === targetNumberOfOccurrences) {
+      return hexColorSequence[i];
+    }
+  }
+
+  return hexColorSequence[overflow];
+};
+export const getNextClassColor = (attibutedColors: string[]): string => {
```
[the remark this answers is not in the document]
Jordanlelay: Solved

typescript/utils/class-color-generator.ts
Jordanlelay: Not sure why, I kept the same definition

typescript/utils/class-color-generator.ts
Jordanlelay: Done

typescript/utils/class-color-generator.ts
valerian: I agree that the none color does not really belong here : this file is for generating colors for existing classes that will be stored, but the none color is not for a class, it's more in the UI design category, part of the site's style.

Jordanlelay: I moved it

valerian (changes requested):
I think it would make more sense to move it in `typescript/web/src/theme.ts`, as part of the [Chakra theme](https://chakra-ui.com/docs/theming/customize-theme), so when using it in components you could just use its theme path string in place of the hex code string without needing to import it explicitly.

I would like to have a second opinion from someone else on this though.

Jordanlelay:
> I think it would make more sense to move it in `typescript/web/src/theme.ts`, as part of the [Chakra theme](https://chakra-ui.com/docs/theming/customize-theme), so when using it in components you could just use its theme path string in place of the hex code string without needing to import it explicitly.
> 
> I would like to have a second opinion from someone else on this though.

Maybe you should create a dedicated issue for that because it's not related with this one, what do you think ?

vincent-lecrubier-skydio:
> I think it would make more sense to move it in `typescript/web/src/theme.ts`, as part of the [Chakra theme](https://chakra-ui.com/docs/theming/customize-theme), so when using it in components you could just use its theme path string in place of the hex code string without needing to import it explicitly.
> 
> I would like to have a second opinion from someone else on this though.

I'd say probably not, because label colors are a functional thing, not a themeing thing. They get written to the DB, get written to the exported files etc.. Also we might want to user the color generator in server side code in the DB, for example in COCO importers or similar

valerian:
> > I think it would make more sense to move it in `typescript/web/src/theme.ts`, as part of the [Chakra theme](https://chakra-ui.com/docs/theming/customize-theme), so when using it in components you could just use its theme path string in place of the hex code string without needing to import it explicitly.
> > I would like to have a second opinion from someone else on this though.
> 
> I'd say probably not, because label colors are a functional thing, not a themeing thing. They get written to the DB, get written to the exported files etc.. Also we might want to user the color generator in server side code in the DB, for example in COCO importers or similar

My request is only regarding the constant defining the "neutral" color to display when a label has no class.

The code for attributing a new color to a class is indeed in the right location as a shared util.

vincent-lecrubier-skydio:
Ah yes then I don’t know !

bbenoist:
Why not in `typescript/web/src/theme.ts` ?

valerian (approved):
Approved, but let's make a small issue about including the constant in the theme

Jordanlelay:
> Approved, but let's make a small issue about including the constant in the theme

Ok I will create it

Issue #555, opened 2021-11-04
Update tutorial images for IOG + workspace


nicOwlas:
Mockups are [here](https://www.figma.com/file/4J9xyfF5v4B98MuKVO0Y8V/LabelFlow-mockups?node-id=3953%3A13826):
![image](https://user-images.githubusercontent.com/16015322/146915906-261c9550-d1b5-4742-bedc-17526c8fbfbf.png)

## Tasks
- [x] Create dataset with compressed files here (keep images in alphabetical order): [Resized.zip](https://github.com/labelflow/labelflow/files/7752193/Resized.zip)
- [x] Use the labels and class of the existing tutorial dataset

nicOwlas:
Regarding the low visibility of the `None` class labels on some images. Lets keep it simple and keep it like this. I'll edit the tutorial images to avoid greyish background.

What some other platforms are doing:
- There is no none class, you need to create/pick a class when you create a label. You can select a color for this label. You'll eventually run into some low visibility problems on images where the background has similar color to the label color.
- There is a side bar which displays the existing labels on the image and you can select it by clicking on the sidebar. When you click on a label in the list, it gets selected in the labeling view.
- You can navigate between labels by clicking on tab. The view is centered on the selected label.

nicOwlas:
Optimize image size in sqoosh

valerian:
One feature that users might like, to improve the labels visibility in general, would be a toggle button to make the image like 50% darker, it would also be easier for the eyes in dark mode. Does anyone know if other labelling softwares have a similar feature?

nicOwlas:
> One feature that users might like, to improve the labels visibility in general, would be a toggle button to make the image like 50% darker, it would also be easier for the eyes in dark mode. Does anyone know if other labelling softwares have a similar feature?

Supervise.ly has something similar. Not sure for the others. In this issue we'll stick to changing the tutorial images to avoid gray above white.
![image](https://user-images.githubusercontent.com/16015322/146913145-dfb3b493-9523-41e1-aafe-d77b8d5faddd.png)

Jordanlelay:
I had to modify the labels on the horses image because of the resize, they were not placed correctly

nicOwlas:
> I had to modify the labels on the horses image because of the resize, they were not placed correctly

Thanks for that, I was not sure if the size was 2500 of 3000 pixel wide on the original tutorial images.

Jordanlelay:
solved by #677

PR #677, opened 2021-12-22
Update tutorial images and labels

Jordanlelay:
# Feature

## Work performed

- Update tutorial dataset images
- Update the labels on tutorial image

## Results

There is now 6 images in the tutorial dataset

## Problems encountered

Because the new images are not at the same scale, I had to update the labels positions and geometry

## Resolved issues

This PR solves #555


nicOwlas:
All good! 👍

PR #683, opened 2021-12-23
Remove cross-projects relative imports

bbenoist:
# Feature

## Work performed

They were some cross-projects relative imports. I've changed them to `@labelflow/*`

## Results

No more imports using relative path between the projects of this monorepo

## Problems encountered

No

## Caveats

None

## Resolved issues

Needed by #615 

## Newly raised issues

None

PR #685, opened 2021-12-23
Generate Prisma client outside of node_modules

bbenoist:
Depends on #683 

## Work performed

By default, Prisma generates the client in `node_modules/.prisma`.

To prevent future circular-project dependencies and make sure that the dependency chain remains untouched, I've changed it to `typescript/db/src/__generated__/client`._

## Results

* Prisma client files are generated in `typescript/db/src/__generated__/client`
* `PrismaClient` is now imported from `@labelflow/db` (or a relative path if code is in `db`) instead of `@prisma/client`

## Problems encountered

Generated files must be ignored by Git, ESLint and Prettier (+Docker with on-premise)

## Caveats

Adds few more ignore lines in the ignore files

## Resolved issues

Needed by #615 

## Newly raised issues

None


typescript/db/src/repository/index.ts
```diff
@@ -1,4 +1,4 @@
-import { Prisma } from "@prisma/client";
+import { Prisma } from "../__generated__";
```
nacho-carnicero, line 1: I get an error from eslint here: `../__generated__ import should occur after import of slugify `

typescript/db/src/repository/label.ts
```diff
@@ -1,4 +1,4 @@
-import { Prisma } from "@prisma/client";
+import { Prisma } from "../__generated__";
```
nacho-carnicero, line 1: Same here, the relative import should come after the absolute imports

typescript/db/src/repository/workspace.ts
```diff
@@ -1,4 +1,4 @@
-import { Prisma, WorkspacePlan, UserRole } from "@prisma/client";
+import { Prisma, WorkspacePlan, UserRole } from "../__generated__";
```
nacho-carnicero, line 1: Move the relative import to after the absolute imports

typescript/db/src/resolvers/user.ts
```diff
@@ -1,4 +1,4 @@
-import { Prisma } from "@prisma/client";
+import { Prisma } from "../__generated__";
```
nacho-carnicero, line 1: Move the relative import

typescript/db/src/resolvers/workspace.ts
```diff
@@ -1,4 +1,4 @@
-import { Prisma } from "@prisma/client";
+import { Prisma } from "../__generated__";
```
nacho-carnicero, line 1: Move the relative import

nacho-carnicero:
Apart from those small comments that should make the lint job pass this looks good to me, ok to merge once the tests pass!

Issue #665, opened 2021-12-16
Add spinner on the Label Class page. Investigate on filtering/search not working.


nicOwlas:
- [x] Stress test, behavior with 100, 300, 1000 classes? Works well with 402 classes
- [x] Query the complete list of classes, one unique query
- [x] The entire page is taking some time to load. Add a spinner to show that the data is loading
- [x] Investigate the filtering issues on the search field of the class selection menus.

nicOwlas:
@Jordanlelay @valerian the class list remains fluid even with 402 classes :+1: 

Some issues though (I'll edit the issue correspondingly):
- It takes a few seconds to load the list, we need to add a spinner to show it is loading
- The reordering of the classes remains very usable (small delay at the beginning but nothing blocking)
- Problem: in the labeling view, the search is failing in the class component. The filtering is not working anymore.

Recording

https://user-images.githubusercontent.com/16015322/146932681-24d14723-1402-47a4-804f-67522fc0fe1a.mov

:

Jordanlelay:
@nicOwlas If the search does not work, I think it's because you have several classes with the same name. I have created 400 classes with different names and the search keeps working (it's a bit slow). If you have 2 classes with the same name, all sort of problems start appearing. Because we give React a reference to the list element in the popover to the class by its name, having several times the same name produces errors. You should retry the search with 400 uniques elements if the problem persist, I will investigate more.

I'm assuming that 2 classes **cannot** have identical names inside the same dataset

PR #692, opened 2021-12-24
Update for using reusable custom spinner

Jordanlelay:
# Feature

## Work performed

- Add a reusable spinner component
- Add a spinner in dataset-class-table

## Resolved issues

Solves #665


typescript/web/src/components/spinner/index.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+type SpinnerProps = {
+  id?: string | undefined;
+  color?: string | undefined;
+  ariaLabel?: string | undefined;
+  size?: ThemingProps<"Spinner">["size"] | undefined;
+  visibility?: "visible" | "hidden" | undefined;
+  key?: string | undefined;
+  ref?: React.Ref<HTMLDivElement> | undefined;
+};
```
bbenoist, line 12: Why don't you directly reuse `SpinnerProps` from Chakra?

typescript/web/src/components/spinner/index.tsx
```diff
@@ -0,0 +1,35 @@
...
+export const Spinner = ({
+  id = undefined,
+  color = "black",
+  ariaLabel = "",
+  size = "xl",
+  visibility = "visible",
+  key = undefined,
+  ref = undefined,
```
bbenoist, line 20: Are all there default props really needed?

typescript/web/src/components/spinner/index.tsx
```diff
@@ -0,0 +1,35 @@
...
+}: SpinnerProps) => {
+  return (
+    <ChakraSpinner
+      id={id}
+      aria-label={ariaLabel}
+      thickness="4px"
```
bbenoist, line 26: Do we really need to override Chakra default thickness?

typescript/web/src/pages/[workspaceSlug]/datasets/index.tsx
```diff
@@ -58,13 +59,7 @@ const LoadingCard = () => (
...
       justify="center"
     >
-      <Spinner
-        thickness="4px"
-        speed="0.65s"
-        emptyColor="gray.200"
-        color="#31CECA"
-        size="xl"
-      />
+      <Spinner color="#31CECA" />
```
Comment (bbenoist) [inline typescript/web/src/pages/[workspaceSlug]/datasets/index.tsx:62]: While you're modifying this block, can you make sure that this color cannot come from the theme?

typescript/web/src/components/spinner/index.tsx
```diff
@@ -0,0 +1,35 @@
+import React from "react";
+import { Spinner as ChakraSpinner, ThemingProps } from "@chakra-ui/react";
+
+type SpinnerProps = {
+  id?: string | undefined;
```
bbenoist, line 5: It's preferable to choose between `?` and `| undefined` (`?` if the property is optional, `| undefined` if you want to force the property to be set in any case).
There is no need for both (and some stricter configuration would prevent you to do both).

typescript/web/src/pages/[workspaceSlug]/datasets/index.tsx
Comment (Jordanlelay) [reply typescript/web/src/pages/[workspaceSlug]/datasets/index.tsx]: Right, I changed it

typescript/web/src/components/spinner/index.tsx
Jordanlelay: No, not necessarily. I reuse the value of Nico's example, but I guess default value will work just fine.

typescript/web/src/components/spinner/index.tsx
Jordanlelay: Fixed

typescript/web/src/components/spinner/index.tsx
Jordanlelay: Done

typescript/web/src/components/spinner/index.tsx
Jordanlelay: Fixed

bbenoist (approved):

Issue #660, opened 2021-12-16
Paginate the image gallery tab


nicOwlas:
## Objective:
Paginate the image gallery tab to have a smooth loading even on large datasets

## Tasks:
  - [x] Display pages of X elements with X = 50/100/1000 is a good compromise to avoid the complexity introduced by infinite scrolling for now.
  - [x] Page selector is similar to the image selector on the labeling view
  - [x] Segmented controls are used to select the number of items per page
  - [x] We display the index of items displayed. It does not correspond to the items selected. Page 10 with 50 images per page we would display `501-550 of 720`
  - [x] The bottom bar does not scroll, it's always visible. This allows to not have to scroll to the bottom of the page or back up to change the page.

## Mockup ([link](https://www.figma.com/file/4J9xyfF5v4B98MuKVO0Y8V/LabelFlow-mockups?node-id=4008%3A13848))
![image](https://user-images.githubusercontent.com/16015322/147340214-9b2b6ebf-10a3-441c-96fd-0ddb793a7fb3.png)

nicOwlas:
Regarding pagination of images: [see conversation here](https://labelflow.slack.com/archives/C022Q91MGCX/p1639667305015900)

nicOwlas:
Mockup for validation (issue description updated) ([link](https://www.figma.com/file/4J9xyfF5v4B98MuKVO0Y8V/LabelFlow-mockups?node-id=3962%3A14943))
![image](https://user-images.githubusercontent.com/16015322/146969081-6203b934-cc12-4254-8a90-82a82ff2b53d.png)

valerian:
@nicOwlas do we want pages to have their own URL?

nicOwlas:
> @nicOwlas do we want pages to have their own URL?

Not for now. I can't identify a use case where it would be useful. Also the other platforms mentioned[ in this thread](https://labelflow.slack.com/archives/C022Q91MGCX/p1639667305015900) are not doing it.

crubier:
Me when I see the pagination interface

![5dbpp5](https://user-images.githubusercontent.com/2954572/147264566-8a03fcc0-8249-41f4-a625-ba0b4cbaae02.jpg)

nicOwlas:
> Me when I see the pagination interface

Opened to suggestion

nicOwlas:
Some alternative designs

- 
[...]

PR #693, opened 2021-12-24
Paginate dataset image gallery

valerian:
# Feature

## Work performed

- Made the `ToggleButtonGroup` component generic, and changed its default selected background color from gray-200 to gray-300.
- On dataset images tab, made a working pagination interface at the bottom, according to #660 :
  - Page selector, similar to the image selector on the labelling view.
  - Segmented controls to select the number of items per page (50/200/1000).
  - Display the indexes of items displayed, such as "501-550 of 720".
  - Pagination tools are on a bottom bar, always visible (fixed position).

## Caveats

- The pagination toolbar gets covered by the cookie banner

## Resolved issues

Solves #660


nicOwlas:
@valerian I have a 500 error when trying to sign-in. Are you having the same?

nicOwlas:
- When I import images from the image gallery tab. The count of images is updated but not the cards in the gallery.
We should refresh the gallery to load the newly imported images (ideally refresh the gallery only when new elements were added to the page displayed to the user).
image

typescript/web/src/components/import-button/import-images-modal/import-images-modal.tsx
```diff
@@ -51,6 +52,7 @@ export const ImportImagesModal = ({
         fetchPolicy: "network-only",
       });
       client.query({ query: getDatasetsQuery, fetchPolicy: "network-only" });
+      client.refetchQueries({ include: [paginatedImagesQuery] });
```
nacho-carnicero, line 55: This should be removed right?

typescript/web/src/components/pagination-tool/pagination-tool.tsx
```diff
@@ -0,0 +1,134 @@
...
+          />
+        </Tooltip>
+      ) : (
+        <IconButton
+          disabled
+          bg="none"
+          aria-label="No previous page"
+          icon={<RiArrowLeftSLine size="1.5em" />}
+        />
+      )}
```
nacho-carnicero, line 63: Instead of doing a conditional rendering, I'd use the `isDisabled` prop of `Tooltip` and always render both components, disabling them both when `currentPage <=1`

typescript/web/src/components/pagination-tool/pagination-tool.tsx
```diff
@@ -0,0 +1,134 @@
...
+            setPage(inputValue);
+          }}
+        >
+          <NumberInputField
+            rounded={6}
+            onClick={(event) => {
+              (event.target as HTMLInputElement).select();
+            }}
+            textAlign="center"
+            w={`${digitCount * digitsPerRem + 1.5}rem`}
```
nacho-carnicero, line 92: I'd use here a Chakra value for the width, such as `2`, `3` or `4`. This enables to have a consistent design across the different components (same thing below)

typescript/web/src/components/pagination-tool/pagination-tool.tsx
```diff
@@ -0,0 +1,134 @@
...
+        mr={3}
+        ml={2}
+        w={`${digitCount * digitsPerRem + 0.5}rem`}
+        textAlign="left"
+        cursor="default"
+        fontSize="sm"
+      >
+        {maxPage}
+      </Text>
+      {currentPage < maxPage ? (
```
nacho-carnicero, line 111: Same thing as before, I'd avoid the conditional rendering by using the `isDisabled` prop of the Tooltip component

typescript/web/src/components/pagination-tool/pagination-tool.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+  useEffect(() => {
+    setInputValue(currentPage);
+  }, [currentPage]);
+
+  const handleKeyPress = (event: any) => {
+    if (event.key === "Enter") {
+      setPage(event.target.value as number);
+    }
+    if (event.key === "Escape" || event.key === "Esc") {
```
nacho-carnicero, line 35: This does not seem to work, at least in mac

typescript/web/src/components/pagination-tool/pagination-tool.tsx
```diff
@@ -0,0 +1,134 @@
...
+            if (!isNaN(valNum)) {
+              setInputValue(clamp(valNum, 1, maxPage));
+            }
+          }}
+          onKeyPress={handleKeyPress}
+          onBlur={() => {
+            setPage(inputValue);
+          }}
+        >
+          <NumberInputField
```
nacho-carnicero, line 86: We can't erase the number completely to set a new number, we have to set the new number and then delete the precedent one. Example: if I'm on page one and I want to type 2 to go to page 2, I need to first type number 2, so for a moment I have the number 12 in there, and then delete the 1

typescript/web/src/components/pagination-toolbar/pagination-toolbar.tsx
```diff
@@ -0,0 +1,100 @@
...
+    setPagesAndItemsPerPages(maxPages, itemsPerPage);
+  }
+
+  return (
+    <Grid
+      pos="fixed"
+      bottom="0px"
+      w="100%"
+      h="44px"
+      zIndex={2}
```
nacho-carnicero, line 50: Why do we need this?

typescript/web/src/components/pagination-toolbar/pagination-toolbar.tsx
```diff
@@ -0,0 +1,100 @@
...
+  if (maxPages && currentPage > maxPages) {
+    setPagesAndItemsPerPages(maxPages, itemsPerPage);
+  }
+
+  return (
+    <Grid
+      pos="fixed"
+      bottom="0px"
+      w="100%"
+      h="44px"
```
nacho-carnicero, line 49: Unless necessary, try to use Chakra values instead of directly pixels (same thing goes for everywhere you're using pixel values)

typescript/web/src/components/toggle-button-group/toggle-button.tsx
```diff
@@ -30,13 +30,13 @@ export const ToggleButton = (props: ToggleButtonProps) => {
...
         <input {...inputProps} aria-labelledby={id} />
-        <IconButton
+        <Button
           as="div"
           id={id}
           color={useColorModeValue("gray.600", "whiteAlpha.700")}
           _checked={{
             color: useColorModeValue("inherit", "whiteAlpha.900"),
-            bg: useColorModeValue("gray.200", "whiteAlpha.300"),
+            bg: useColorModeValue("gray.100", "whiteAlpha.300"),
```
nacho-carnicero, line 39: Why this change?

typescript/web/src/components/pagination-toolbar/pagination-toolbar.tsx
```diff
@@ -0,0 +1,100 @@
...
+        >
+          {Object.keys(itemsPerPageStrToNum).map((itemsPerPageOption) => (
+            <ToggleButton
+              size="sm"
+              key={`imagesPerPageButton-${itemsPerPageOption}`}
+              value={itemsPerPageOption}
+              description={`${itemsPerPageOption} images per page`}
+              pointerEvents="initial"
+              borderColor={mode("gray.100", "whiteAlpha.300")}
+              _checked={{
```
nacho-carnicero, line 74: This is the same as the default values set in the ToggleButton component no? I would say you don't need to pass it

nacho-carnicero:
@valerian I did the review you asked for, apart from some minor stuff it's looking good so far, good job!

typescript/web/src/components/pagination-tool/pagination-tool.tsx
valerian: I have taken this code from the component used for navigating between image in the labelling view, it allows for scaling the number input field to fit the same number of digits as the maximum page number

bbenoist (changes requested):
_:warning: I've not reviewed the Apollo cache parts in details yet._

typescript/web/src/components/pagination-tool/pagination-tool.tsx
valerian: @crubier do you have an answer to that? It seems that you originally coded that part on the [image navigation tool](https://github.com/labelflow/labelflow/blob/752c7389a11e32af6c4086facc2a97e4c0c4c858/typescript/web-app/src/components/image-navigation-tool-bar/image-navigation-tool.tsx), which I partially copied

vincent-lecrubier-skydio: This number only depends on the font family used. We define the font in CSS so there is no reason for this number to change. It does not depend on zoom or font scaling since it is expressed in rem which also scale with zoom or font scaling, canceling each other. 

I found this number by carefully trying various values in storybook, and finding the optimal value that: 

- is not too wide, to avoid having an overly large margin in the box when the number is made of narrow digits (worst case is 11111)
- is not too narrow, to avoid cropping/overflowing the box when the number is made of wide digits (worst case is 88888 or 66666 i think)

The value is precise and 0.6 is too large and 0.5 is too small. 

I don’t think there is an easy way to test that, and anyway you’d be testing the test engine font rendering more than your own code. Also this value is not likely to change often so there is no reason to spend energy testing it really.

vincent-lecrubier-skydio: There are other methods of fitting an input box size to its content but 

- it’s better to only depend on the number of digits and not on the actual exact width of the number, to avoid constantly changing the width of the box every time the number change, which would be bad UX  (here we only change the width when we go from 9 to 10, … , 9999 to 10000 etc..)
- the methods to fit an input to its content often involve complex measurement process that involve DOM operations and CSS which are too costly for the need

typescript/web/src/components/dataset-images-list/image-card.tsx
```diff
@@ -0,0 +1,84 @@
...
+              {name}
+            </Heading>
+            <IconButton
+              icon={<TrashIcon />}
+              aria-label="delete image"
+              isRound
+              size="sm"
+              onClick={(e) => {
+                e.preventDefault();
+                e.stopPropagation();
```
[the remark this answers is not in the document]
valerian: good question, it comes from this https://github.com/labelflow/labelflow/commit/4d595801c18842722a83191deae3827cd09aefca
@nacho-carnicero

typescript/web/src/components/dataset-images-list/images-list.tsx
```diff
@@ -1,149 +1,134 @@
...
 
-const TrashIcon = chakra(HiTrash);
+const imagesPerPageOptions = [50, 200, 1000];
 
 export const ImagesList = ({
   datasetSlug,
   workspaceSlug,
-  images,
+  datasetId,
+  imagesTotalCount,
```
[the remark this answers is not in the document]
valerian: It was to be consistent with the GraphQL name, coming from
```
dataset {
    imagesAggregates {
        totalCount
    }
```

bbenoist (approved):

typescript/web/src/components/pagination/page-input.context.tsx
```diff
@@ -48,7 +48,7 @@ const useHandleKeyDown = (reset: () => void) => {
   const { setPage } = usePagination();
   return (event: KeyboardEvent<HTMLInputElement>) => {
     if (event.key === "Enter") {
-      const newPage = parseInputValue(event.currentTarget.value);
+      const newPage = parseInputValue((event.target as HTMLInputElement).value);
```
bbenoist, line 51: @valerian Sorry. I thought `currentTarget` was defined in this case 😢

typescript/web/src/connectors/apollo-client/cache-config.ts
```diff
@@ -0,0 +1,104 @@
...
+  existing,
+  options: FieldFunctionOptions<QueryImagesArgs, Record<string, any>>
+) => {
+  // A read function should always return undefined if existing is
+  // undefined. Returning undefined signals that the field is
+  // missing from the cache, which instructs Apollo Client to
+  // fetch its value from your GraphQL server.
+  if (existing === undefined) {
+    return undefined;
+  }
```
bbenoist, line 24:
```suggestion
  if (existing === undefined) return undefined;
```

typescript/web/src/connectors/apollo-client/cache-config.ts
```diff
@@ -0,0 +1,104 @@
...
+  options: FieldFunctionOptions<QueryImagesArgs, Record<string, any>>
+) => {
+  // Slicing is necessary because the existing data is
+  // immutable, and frozen in development.
+  const merged = existing
+    ? {
+        images: existing.images.slice(0),
+        totalCount: existing.totalCount,
+      }
+    : { images: [], totalCount: undefined };
```
bbenoist, line 60:
```suggestion
  // Spreading the images array is necessary because the existing data is
  // immutable, and frozen in development
  const merged = {
    images: [...(existing?.images ?? []),
    totalCount: existing?.totalCount
  }
```

typescript/web/src/connectors/apollo-client/cache-config.ts
```diff
@@ -0,0 +1,104 @@
...
+  const offset = options.args?.skip || 0;
+  const limit = options.args?.first || undefined;
+  for (let i = 0; i < incoming.length; i += 1) {
+    merged.images[offset + i] = incoming[i];
+  }
```
bbenoist, line 65: _No really a change-request but rather a recommendation._
For such cases, I'd use [immutability-helper](https://www.npmjs.com/package/immutability-helper) which can, in a single operation, take care of making a deep copy (_but without being subject to object properties changes_) and setting only the updated values in the array. It's been made exactly to cover such cases (_it was created by the React team to help updating deep states_)

typescript/web/src/connectors/apollo-client/cache-config.ts
```diff
@@ -0,0 +1,104 @@
...
+        )
+      : existing.images;
+  if (
+    result.length === 0 ||
+    (result as (Image | undefined)[]).includes(undefined) ||
+    // Invalidate cache if returned result is smaller than the limit
+    // but does not match the expected total count of images in the
+    // dataset
+    (((limit && result.length < limit) || !limit) &&
+      (!existing.totalCount || existing.totalCount !== offset + result.length))
```
bbenoist, line 41: That's what I call a big condition 😅 Can we split it into multiple well-named parts to reduce it's reading complexity?

bbenoist (changes requested):

typescript/web/src/components/pagination/pagination.context.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+const usePaginationState = ({
+  page: defaultPage = 1,
+  perPageOptions = DEFAULT_PER_PAGE_OPTIONS,
+  perPage: defaultPerPage = undefined,
+  itemCount,
+}: Omit<PaginationProps, "children">): PaginationState => {
+  const [pagination, setPagination] = useReducer(paginationReducer, {
+    page: defaultPage,
+    perPage: defaultPerPage ?? perPageOptions[0],
```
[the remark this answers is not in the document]
valerian: I didn't know this was possible in TS/JS !

typescript/web/src/components/pagination/pagination.context.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+export const usePagination = () => useContext(PaginationContext);
+
+const paginationReducer = (
+  prevState: PaginationPayload,
+  nextState: Partial<PaginationPayload>
+): PaginationPayload => ({ ...prevState, ...nextState });
+
+export type PaginationProps = Pick<PaginationState, "itemCount"> &
+  Pick<PaginationState, "perPageOptions"> &
```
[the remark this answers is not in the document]
valerian: My bad, tried `,`, forgot about `|`

typescript/web/src/connectors/apollo-client/cache-config.ts
valerian: I thought we enforced the use of curly brackets ?

bbenoist: I do this (**only for simple one-liner early returns**) since my arrival and never heard it was forbidden 😉

typescript/web/src/connectors/apollo-client/cache-config.ts
valerian: I don't think immutability-helper would be of much help in that situation, in which I need to offset the indexes beforehand anyway.

typescript/web/src/components/dataset-images-list/paginated-images-query.ts
```diff
@@ -16,12 +22,16 @@ export const flushPaginatedImagesCache = async (
   apolloClient: ApolloClient<object>,
   datasetId: string
 ) => {
-  // The custom variable "flush" is only expected by our custom
-  // apollo cache merge function, and will invalidate the cache
-  await apolloClient.writeQuery({
-    query: paginatedImagesQuery,
-    data: { images: [] },
-    variables: { where: { datasetId }, flush: true },
+  const storeFieldName = `images:${JSON.stringify({ where: { datasetId } })}`;
```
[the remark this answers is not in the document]
valerian: Please see this line: https://github.com/labelflow/labelflow/blob/d040e16d76f32a58804850a1f346685b6c16bda3/typescript/web/src/connectors/apollo-client/cache-config.ts#L75

And that:
https://github.com/apollographql/apollo-client/blob/main/src/cache/inmemory/object-canon.ts

And in the current situation there is always only one property, so no risk of sort order issues.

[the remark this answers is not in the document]
valerian: By the way I just realized I can use `canonicalStringify` from `object-canon.ts` instead of `JSON.stringify`, that would be better here.

bbenoist (approved):

PR #695, opened 2021-12-27
Virtualize label class list

### LINKED ISSUE #694: Labelling interface is very slow when there are multiple label classes
**Describe the bug**
The labelling interface gets very slow when there are multiple label classes. The open layers part is fine, but all the business logic gets very slow (selecting a label to change it, select a new drawing tool etc...). 

We should virtualize the list of label classes in both menus (the context menu and the upper left menu).

**To Reproduce**
1. Create a new workspace + dataset
2. Upload an image
3. Create 400 label classes in the dataset
4. Create a label in the image
5. Try to use the labelling interface to create new labels with different tools, see how slow it gets

**Expected behavior**
The interface should not get slower as label classes increase

nacho-carnicero:
# Feature

## Work performed

- Virtualize label class list in labelling interface

## Results

This makes the labelling interface much faster when there's a big number of label classes

## Resolved issues

Closes #694


etiennedupont (approved):
LGTM

PR #697, opened 2021-12-28
Don't skip any Jest test

bbenoist:
# Feature

## Work performed

Some tests were skipped.

## Results

```text
Test Suites: 69 passed, 69 total
Tests:       504 passed, 504 total
Snapshots:   6 passed, 6 total
Time:        129.025 s, estimated 173 s
```

## Problems encountered

I still don't really understand why these tests were skipped.

* It wasn't clear why they were some `it.only`
* I'm sure that we can resolve most the flacky issues.

## Caveats

Flacky tests will be back

## Resolved issues

No more skipped tests

## Newly raised issues

- [ ] Fix most of the flacky tests

### DIFF
--- a/cypress/integration/invitation-manager.online.ts
+++ b/cypress/integration/invitation-manager.online.ts
@@ -178,8 +178,7 @@ describe("Invitation Manager (online)", () => {
     cy.location("pathname").should("eq", "/local/datasets");
   });
 
-  // flacky in the CI
-  it.skip("should open the sign in modal if the user isn't logged in", () => {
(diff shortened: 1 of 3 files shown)


nacho-carnicero:
LGTM, I had very likely forgotten to remove the `only` in those two tests. Regarding the `skip` I wasn't the one doing that but the e2e tests appear to pass now so I'd say it's ok. @RaynalHugo was the one skipping it so maybe he had a good reason for doing it

Jordanlelay (approved):

PR #704, opened 2021-12-29
Disable non-deterministic react-openlayers-fiber stories

RaynalHugo:
## Work performed

- Simply disabled snapshot testing for the non-deterministic stories of react-openlayers-fiber
- Disabled the snapshot test for the WelcomeModal with the countdown
- Added a new story of the WelcomeModal without a countdown. Added snapshot tests for this story
- Make WorkspaceCreationModal and WorkspaceNameInput components accept an `origin` prop only specified on stories.

## Results

To be accepted only once: 
https://www.chromatic.com/build?appId=60a646559841590039471afc&number=2644

## Problems encountered

I am not sure to understand why, but it seems that the UI validation step thought I added new stories:
https://www.chromatic.com/build?appId=60a646559841590039471afc&number=2644

### DIFF
--- a/typescript/web/src/components/workspace-name-input/workspace-name-input.fixtures.tsx
+++ b/typescript/web/src/components/workspace-name-input/workspace-name-input.fixtures.tsx
@@ -8,13 +8,14 @@ import {
 } from "@labelflow/common-resolvers";
 import { Mutation, Query } from "@labelflow/graphql-types";
 import { isEmpty, isNil } from "lodash/fp";
-import { useEffect } from "react";
+import { PropsWithChildren, useEffect } from "react";
 import {
   useWorkspaceNameInput,
   WorkspaceNameInput,
   WorkspaceNameInputProvider,
   WorkspaceNameMessage,
 } from ".";
+import { MockableLocationProvider } from "../../utils/mockable-location";
 import { WORKSPACE_EXISTS_QUERY } from "./workspace-exists.query";
 import { WorkspaceNameMessageProps } from "./workspace-name-message";
 
@@ -41,6 +42,7 @@ export type TestComponentProps = Partial<WorkspaceNameMessageProps> & {
   defaultName?: string;
   graphqlMocks?: ApolloResponse<Partial<Query | Mutation>>[];
   storybook?: boolean;
+  origin?: string;
 };
 
 const NameObserver = ({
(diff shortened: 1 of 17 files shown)


typescript/web/src/components/workspace-name-input/workspace-name-input.fixtures.tsx
```diff
@@ -66,7 +67,11 @@ export const TestComponent = ({
     <WorkspaceNameInputProvider defaultName={defaultName}>
       <NameObserver name={name} storybook={storybook} />
       <WorkspaceNameInput />
-      <WorkspaceNameMessage hideError={hideError} {...messageProps} />
+      <WorkspaceNameMessage
+        hideError={hideError}
+        {...messageProps}
+        origin={storybook ? "http://localhost" : undefined}
```
bbenoist, line 73: There is `localhost` here but `labelflow.ai` below. Why?

typescript/web/src/components/workspace-switcher/create-workspace-modal/__stories__/create-workspace-modal.stories.tsx
```diff
@@ -21,7 +21,11 @@ const Template = () => (
     ]}
   >
     <div>
-      <CreateWorkspaceModal isOpen onClose={console.log} />
+      <CreateWorkspaceModal
+        isOpen
+        onClose={console.log}
+        origin="https://labelflow.ai"
```
bbenoist, line 27: Here ⬆️

bbenoist (commented):
Apart from my small comment, you wouldn't be surprised that I don't really like how we spread `origin` across the components.

Can't you just mock `globalThis.location.origin`?

typescript/web/src/components/workspace-name-input/workspace-name-message.tsx
```diff
@@ -17,10 +18,10 @@ const useUrlMessage = (
   isEditing: boolean | undefined
 ): string | undefined => {
   const { slug } = useWorkspaceNameInput();
-  const hideUrl =
-    !isEmpty(error) || isEmpty(slug) || isEmpty(globalThis.location);
+  const { origin } = useMockableLocation();
```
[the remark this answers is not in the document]
bbenoist: While would I agreed with this before working on this issue, the thing is that:
* Storybook has no way to override the hostname used by its stories via Location (like `siteUrl` for Jest tests).
* We can't easily override Location only for an inner React component and its callbacks.
* `globalThis.location` and `window.location` are used at other places in the code and possibly by some dependencies.
* It's returned values don't trigger re-renders when they're changed as you'd expect with a regular `useLocation` hook (e.g. [react-router version](https://v5.reactrouter.com/web/api/Hooks/uselocation)).

As such, this hooks and its provider are really only useful when you need to use an instance Location which can me mocked. You don't really need it in other cases.

Also, the component still don't have the responsibility to know whether it is mocked or not (that's why I used `mockable` instead of `mocked` in the name). It's under the responsibility of the component which has declared the correct provider (the step wrapper or the main app entrypoint).

valerian (approved):

PR #709, opened 2021-12-29
Make sure that at-least one authentication provider is defined

bbenoist:
## Work performed

I've made sure that at-least one of the authentication providers is correctly configured when starting the server.

## Results

Server will throw an error if none of the auth providers are configured.

## Problems encountered

No

## Caveats

We must find a solution which doesn't rely on any external service (username/password?)

## Resolved issues

Helps with #615 

## Newly raised issues

To create:  Find a solution which doesn't rely on any external service (username/password?)


typescript/web/src/pages/api/auth/[...nextauth].ts
```diff
@@ -30,24 +31,66 @@ if (!globalThis.prismaInstance) {
...
+    "GOOGLE_CLIENT_ID",
+    "GOOGLE_CLIENT_SECRET",
+    GoogleProvider
+  );
+};
+
+const getGitHubProvider = () => {
+  return getOAuthProvider(
+    "GOOGLE_CLIENT_ID",
+    "GOOGLE_CLIENT_SECRET",
```
Comment (Jordanlelay) [inline typescript/web/src/pages/api/auth/[...nextauth].ts:75]:
```suggestion
    "GITHUB_CLIENT_ID",
    "GITHUB_CLIENT_SECRET",
```

Jordanlelay (changes requested):

typescript/web/src/pages/api/auth/[...nextauth].ts
Comment (bbenoist) [reply typescript/web/src/pages/api/auth/[...nextauth].ts:75]: 😱 Thank you @Jordanlelay

valerian:
E2E tests fails with `Error: No authentication provider has been configured`
Cypress runs with `NODE_ENV="production"`: should we change that?

PR #713, opened 2021-12-30
Move some functions into @labelflow/utils

bbenoist:
## Work performed

As @Jordanlelay has introduced the utils package in the project, I've moved some low-level functions we had to the package.

## Results

The package now includes:
* sleep
* timeout
* reorderArray

## Problems encountered

It wasn't funny to mock time with jest and find a way to wait for all promises to be executed (implemented in `flushPromises`).

## Caveats

?

## Resolved issues

Somewhat helps with #551 

## Newly raised issues

No


typescript/utils/src/__tests__/sleep.test.ts
```diff
@@ -0,0 +1,14 @@
+import { sleep } from "../sleep";
+
+jest.useFakeTimers();
+
+describe("sleep", () => {
+  it("waits for the specified amount of time", async () => {
+    const resolve = jest.fn();
+    const reject = jest.fn();
+    sleep(100).then(resolve).catch(reject);
+    jest.advanceTimersByTime(100);
```
valerian, line 10:
```suggestion
    jest.advanceTimersByTime(50);
    expect(resolve).not.toHaveBeenCalled();
    jest.advanceTimersByTime(60);
    expect(resolve).toHaveBeenCalled();
```

typescript/utils/src/__tests__/sleep.test.ts
```diff
@@ -0,0 +1,14 @@
...
+    await Promise.resolve();
+    expect(resolve).toHaveBeenCalled();
```
valerian, line 12:
```suggestion
```

valerian (changes requested):

typescript/utils/src/__tests__/sleep.test.ts
valerian: it seems that my suggestion doesn't work as expected after all, but this might help: https://stackoverflow.com/questions/66391541/jest-advancetimersbytime-doesnt-work-when-i-try-to-test-my-retry-util-function/66415212#66415212

bbenoist: It doesn't work as-is because you've set the second value to 60 whereas the sleep is 100 😉

bbenoist: And because it's placed above the `await Promise.resolve()` line which forces the promise engine to have a new tick and resolve the Promise returned by sleep (can you feel the headache I had while implementing this test now? 😝)

typescript/utils/src/__tests__/sleep.test.ts
bbenoist: ...or not (for the reasons above) 😉

typescript/utils/src/__tests__/sleep.test.ts
valerian, line 10:
```suggestion
    jest.advanceTimersByTime(50);
    await Promise.resolve();
    expect(resolve).not.toHaveBeenCalled();
    jest.advanceTimersByTime(50);
```

This works as expected, and now tests both before and after the sleep delay.

valerian (changes requested):

valerian (approved):

typescript/utils/src/__tests__/sleep.test.ts
bbenoist: Nice ❤️ Thank you!

Issue #712, opened 2021-12-30
Create a function decomposing apollo errors to eventually display its details to the user


bbenoist:

nicOwlas:
@bbenoist can we close?

bbenoist:
Well, I've built basic implementations in [`getApolloErrorMessage`](https://github.com/labelflow/labelflow/blob/cfb956f89fa35e8bcdf97bebf2810ec727911a3a/typescript/web/src/utils/get-apollo-error-message.tsx#L24) and [`useApolloErrorToast`](https://github.com/labelflow/labelflow/blob/3998bc2b54575ed861c317731ef7cf9eb971cc14/typescript/web/src/components/toast.ts#L39) but it still needs to be reused elsewhere than in `CreateWorkspaceModal` and `WorkspaceNameInput`.

It also needs to catch more errors (I've mostly made it work for #710 and #714) and it still lacks an official way of throwing errors from the server.

Anyway, since we're working on #741 , things could improve drastically and without much efforts at some point but we'll need to go some steps further before then.

As such, I'd say that we could rename this issue with a more general title such as `Improve error handling` and wait for #741 to complete before we could to start thinking again about the issue.

Issue #710, opened 2021-12-30
Users do not get an error when creating a workspace with an already existing Slug


nicOwlas:
## Describe the bug

Users does not get an error when creating a workspace with an already existing Slug

## To Reproduce

1.  Sign in and create a workspace named "Demo"
2.  You will get an error 400
3. 
4. 

## Expected behavior

Alert the user, this workspace name is already taken.

## Screenshots

![image](https://user-images.githubusercontent.com/16015322/147728521-7c6fa433-eee6-41da-b0fd-f745fa5eac25.png)

![image](https://user-images.githubusercontent.com/16015322/147728429-1f959fe8-687a-4c93-8639-fa12803d65e3.png)

## Device information

 

- **Device:** 
- **OS:** 
- **Browser:** 
- **Version:**

bbenoist:
@nicOwlas Actually, it's already verifying that the slug exists (you can see the details in the console log).
But, as we don't have a standard way to extract the ApolloError message yet, I couldn't display a more meaningful error at the time of writing #667 
This is the problem I've told you duriing one of tthe DSMs. We've created #712 (which I've just converted to an issue) to track this. Would you like me to work on #712 directly?

nicOwlas:
I've see the console details. Hugo had created a resolver `doesWorkspaceSlugExists` (or something like it) which returns a boolean, as an easy fix, can't you use it to display a more meaningful error if it returns `true`?

bbenoist:
This is what I've already used (I've even renamed it `workspaceExists` to get a more obvious name 😉 ).

You had this error message because you've submitted the form before the background call to `workspaceExists` could end (as the verification is debounced and non-blocking).
If you wait before submitting, you'll see that the workspace name verification is working as expected and displays the correct error message.

In other words, the `workspaceExists` verification on the client-side doesn't really prevent you from submitting the form if it is not finished yet (simpler verifications like forbidding special characters are immediate BTW).
If the `workspaceExists` has time to finish before the form could be submitted, it will block the form from being submitted and display a meaningful error message (which is the exact same returned by `createWorkspace` in the `ApolloError` details).
But, if it didn't had enough time before the end-user could submit the form, we delegate the verification responsibility to the server which is in anyway the real source of truth.
This mechanism is made so that the golden path remains as fast as possible. We have actually no reason to wait for `workspaceExists` since it's already done by `createWorkspace`.

nicOwlas:
[...]

PR #714, opened 2021-12-30
Display a more meaningful error message in the app when createWorkspace fails

bbenoist:
## Work performed

I've added `getApolloErrorMessage` which tries to find the more meaningful error as possible so that it can be displayed to the user.

## Results

image

## Problems encountered

* Prisma errors are not fully typed
* Apollo errors are not fully typed

## Caveats

We might miss additional errors.

## Resolved issues

Closing #710 
Also related with #712 

## Newly raised issues

No


valerian (approved):
Seems ok

Issue #705, opened 2021-12-29
Multiple classes are created at once


nicOwlas:
## To Reproduce

1.  Go to an online workspace
2.  Create a new dataset
3.  Create a new class by clicking on `+ New Class`
4.  Multiple classes have been created and with the same colors (2 on the screen recording below but I had 5 on another trial)

## Expected behavior

One unique class should be created.

## Screenshots

https://user-images.githubusercontent.com/16015322/147680495-a71c9ccd-ff3d-4386-a99d-ff1efce65ad8.mov

## Device information

 

- **Device:** 
- **OS:** 
- **Browser:** 
- **Version:**

PR #715, opened 2021-12-30
Fix bug creating duplicate label classes from modal

valerian:
## Work performed

- Prevented class creation modal form from being submitted again while a query to create a new class is still ongoing.

## Results

- It shouldn't be possible anymore to create multiple classes with the same name at once

## Resolved issues

Solves #705

### DIFF
--- a/typescript/web/src/components/dataset-classes/upsert-class-modal/upsert-class-modal.tsx
+++ b/typescript/web/src/components/dataset-classes/upsert-class-modal/upsert-class-modal.tsx
@@ -24,7 +24,8 @@ const useCreateClass = (
   className: string,
   classColor: string | undefined,
   onClose: () => void,
-  setErrorMessage: (message: string) => void
+  setErrorMessage: (message: string) => void,
+  setPendingStatus: (isPending: boolean) => void
 ) => {
   const workspaceSlug = useRouter()?.query?.workspaceSlug as string | undefined;
   const updateLabelClass = useUpdateLabelClass(classId, className, classColor);
@@ -39,6 +40,7 @@ const useCreateClass = (
       event.preventDefault();
       if (isEmpty(className)) return;
       try {
+        setPendingStatus(true);
         if (!isNil(classId)) {
           await updateLabelClass();
         } else {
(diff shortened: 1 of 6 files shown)


typescript/web/src/components/dataset-classes/upsert-class-modal/upsert-class-modal.tsx
```diff
@@ -112,6 +117,8 @@ const useModalState = ({ isOpen, onClose }: UpsertClassModalProps) => {
   const { editClass, datasetId, datasetSlug } = useDatasetClasses();
   const [classNameInputValue, setClassNameInputValue] = useState<string>("");
   const [errorMessage, setErrorMessage] = useState<string>("");
+  const [isClassCreationPending, setIsClassCreationPending] =
+    useState<boolean>(false);
```
bbenoist, line 121: Using a dedicated state would trigger additional re-renders which are avoidable by relying on the `loading` variable which is returned by `useMutation`.
If the `createLabelClass` and `updateLabelClass` (whose loading state should have been handled too) don't use `useQuery` yet, just change the code to use this hook or find an alternative based on `client.mutate`.

bbenoist (changes requested):

typescript/web/src/components/dataset-classes/upsert-class-modal/upsert-class-modal.tsx
valerian: Refactoring this code to properly use `useMutation` instead of `ApolloClient.mutate` would require quite a lot of work in its current state, I suggest we:
- merge this pull request as a quick fix for now.
- create a new issue to ask for a proper refactor of the class upsert code from the class list page.

bbenoist: It's not that long actually, I've done in in ~30mn. See the last commit I've just pushed 😉

bbenoist (approved):

typescript/web/src/components/dataset-classes/upsert-class-modal/upsert-class-modal.tsx
valerian: I see, if I did it myself, being not too familiar with react hooks yet, I would have spent quite some time checking that I didn't do anything wrong though :P

PR #717, opened 2021-12-30
Use JWT session tokens

### LINKED ISSUE #716: Migrate from `getSession` to `JWT`
## Objective
Use jwt to avoid calling the DB to check the session token is valid

## Tasks

- [x] Configure next-auth to use jwt ([file here](https://github.com/labelflow/labelflow/blob/main/typescript/web/src/pages/api/auth/%5B...nextauth%5D.ts))
- [ ] Investigate and replace the code where needed

bbenoist:
## Work performed

* Enabled the JWT session in next-auth config
* Upgrade next-auth to version 4.0.3

## Results

Sessions should now use JWT to store the authentication information. Which should reduce the amount of DB queries.

## Problems encountered

* As the `user` argument of the next-auth `session` callback is not defined when using JWT, I had to rely on `token.sub` instead of `user.id`. I've made it so we could still disable JWT without having to change the callback.

## Caveats

* ⚠️ Requires `JWT_SECRET` environment variable to be defined on production. (Must be generated with `openssl rand -base64 32`)
* ⚠️ Existing user sessions will be revoked

## Resolved issues

Closes #716 

## Newly raised issues

No


cypress/fixtures/jwt.ts
```diff
@@ -0,0 +1,31 @@
+import { User } from "@prisma/client";
+// eslint-disable-next-line import/no-extraneous-dependencies
+import { encode as jwtEncode } from "next-auth/jwt";
```
valerian, line 3:
```suggestion
import { encode as jwtEncode, JWT } from "next-auth/jwt";
```

cypress/fixtures/jwt.ts
```diff
@@ -0,0 +1,31 @@
...
+import { v4 as uuid } from "uuid";
+
+const createJwtToken = (user: User) => {
```
valerian, line 6:
```suggestion
const createJwtToken = (user: User): JWT => {
```

nicOwlas:
@bbenoist I'm unable to sign-in on the JWT released version. I get an error screen after clicking the send magic link button when trying to login with email. Recording:
Screenshot 2022-01-06 at 09 15 13 1

bbenoist:
@nicOwlas Thank you, that's what i was afraid of. I'll look at it!

bbenoist:
@nicOwlas I've just forgot to add our JWT secret (a unique cryptographic key used to sign the tokens) in vercel. It should be OK once Vercel has finished redeploying the app :)
I've used a different one in production so that we reduce the chances of being leaked.

nicOwlas:
Working great now 👍

valerian (approved):

PR #732, opened 2022-01-04
Trim white spaces around email addresses

nicOwlas:
## Work performed

trim white spaces around email addresses

## Results

An email address can now have leading or trailing white spaces around it and still be valid as it is trimmed prior to verification.

### DIFF
--- a/typescript/web/src/components/members/new-member-modal.tsx
+++ b/typescript/web/src/components/members/new-member-modal.tsx
@@ -132,7 +132,7 @@ user2@example.com
                     : ""
                 } ${
                   hasInvalidEmails
-                    ? "\nAt lease one email format is invalid. Write one email address per line."
+                    ? "\nAt lease one email format is invalid. Emails should be separated by , ; or spaces."
                     : ""
                 }`}
               </Text>
--- a/typescript/web/src/components/members/new-member-modal.tsx
+++ b/typescript/web/src/components/members/new-member-modal.tsx
@@ -61,7 +61,6 @@ export const NewMemberModal = ({
   };
   const emails = value
     .split(/[,;\s]/)
-    .map((email) => email.trim())
     .filter((email) => email !== "");
   const hasInvalidEmails = emails.some((email) => !validateEmail(email));
 
--- a/typescript/web/src/components/members/new-member-modal.tsx
+++ b/typescript/web/src/components/members/new-member-modal.tsx
@@ -1,34 +1,38 @@
-import { useState, ChangeEvent } from "react";
 import {
-  Flex,
   Alert,
-  AlertIcon,
   AlertDescription,
-  VStack,
+  AlertIcon,
+  Button,
+  Flex,
+  Heading,
+  Link,
+  Modal,
+  ModalBody,
+  ModalCloseButton,
+  ModalContent,
+  ModalFooter,
+  ModalHeader,
+  ModalOverlay,
   Text,
   Textarea,
-  Modal,
-  ModalOverlay,
-  ModalContent,
-  ModalCloseButton,
-  ModalFooter,
-  Button,
-  Heading,
-  ModalHeader,
-  ModalBody,
   useColorModeValue as mode,
-  Link,
   useToast,
+  VStack,
 } from "@chakra-ui/react";
+import { InvitationResult, MembershipRole } from "@labelflow/graphql-types";
+import { isEmpty } from "lodash/fp";
 import { useRouter } from "next/router";
-import { MembershipRole, InvitationResult } from "@labelflow/graphql-types";
-
+import { ChangeEvent, useState } from "react";
+import { validateEmail } from "../../utils/validate-email";
 import { RoleSelection } from "./role-selection";
 import { InviteMember } from "./types";
-import { validateEmail } from "../../utils/validate-email";
 
 const maxNumberOfEmails = 20;
 type EmailStatuses = Record<InvitationResult, string[]>;
+
+const parseEmails = (text: string): string[] => {
+  return text.split(/[,;\s]/).filter((email) => !isEmpty(email));
+};
 
 const summarizeEmailList = (emailList: string[]): string => {
   if (emailList.length === 1) {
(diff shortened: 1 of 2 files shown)


typescript/web/src/components/members/new-member-modal.tsx
```diff
@@ -59,7 +59,10 @@ export const NewMemberModal = ({
     const inputValue = e.target.value;
     setValue(inputValue);
   };
-  const emails = value.split("\n").filter((email) => email !== "");
+  const emails = value
+    .split("\n")
+    .filter((email) => email !== "")
+    .map((email) => email.trim());
```
bbenoist, line 65: The map should be before the filter so that values with just spaces will get removed.

typescript/web/src/components/members/new-member-modal.tsx
```diff
@@ -59,7 +59,10 @@ export const NewMemberModal = ({
     const inputValue = e.target.value;
     setValue(inputValue);
   };
-  const emails = value.split("\n").filter((email) => email !== "");
+  const emails = value
+    .split("\n")
```
bbenoist, line 63: By using `.split(/[\s,;]+/)` (regex to be verified) you could remove the need to `.map` and also allow commas and semicolons.

bbenoist (changes requested):

typescript/web/src/components/members/new-member-modal.tsx
nicOwlas: Done

typescript/web/src/components/members/new-member-modal.tsx
nicOwlas: Done too. I checked the regex.

typescript/web/src/components/members/new-member-modal.tsx
```diff
@@ -59,7 +59,10 @@ export const NewMemberModal = ({
     const inputValue = e.target.value;
     setValue(inputValue);
   };
-  const emails = value.split("\n").filter((email) => email !== "");
+  const emails = value
+    .split(/[,;\s]/)
+    .map((email) => email.trim())
```
bbenoist, line 64: You don't need to trim anymore since the regex splits by spaces 😉 

```suggestion
```

typescript/web/src/components/members/new-member-modal.tsx
```diff
@@ -129,7 +132,7 @@ user2@example.com
                     : ""
                 } ${
                   hasInvalidEmails
-                    ? "\nAt lease one email format is invalid. Write one email address per line."
+                    ? "\nAt lease one email format is invalid. Emails should be separated by , ; or spaces."
```
bbenoist, line 135: [Thanks to `\s` in the regex ](https://regexone.com/lesson/whitespaces), it's still valid (and more readable with the textbox we use) to use new lines.

Also, there is a typo, it's `At-least`, not `At lease`.

```suggestion
                    ? "\nAt least one email address is invalid. Addresses can be separated them by line-breaks, spaces, commas or semicolons."
```

bbenoist (changes requested):

typescript/web/src/components/members/new-member-modal.tsx
valerian: "line-breaks" may be a bit too technical no, maybe "new lines"?

bbenoist: @valerian That's what I was thinking too before realising that it's actually line-breaks, not new-lines.
I agree however that I'd be clearer and abusing the `new line` words has the advantage that everyone will understand. Let's use it 👍 🚀

typescript/web/src/components/members/new-member-modal.tsx
```diff
@@ -1,35 +1,39 @@
...
+import { InvitationResult, MembershipRole } from "@labelflow/graphql-types";
+import { isEmpty } from "lodash/fp";
 import { useRouter } from "next/router";
-import { MembershipRole, InvitationResult } from "@labelflow/graphql-types";
-
+import { ChangeEvent, useState } from "react";
+import { validateEmail } from "../../utils/validate-email";
 import { RoleSelection } from "./role-selection";
 import { InviteMember } from "./types";
-import { validateEmail } from "../../utils/validate-email";
```
bbenoist, line 28: @nicOwlas Apart from the new `isEmpty` import from `lodash/fp`, the changes above are only caused by the "Organize imports" feature of VS Code which lets me gain hundreds of cumulated hours reorganizing imports manually by doing it automatically.

_It might be nice to enable it by default in the future (and ignore it in the rare places where the import order actually matters) so that the whole team could gain some dev time too._

typescript/web/src/components/members/new-member-modal.tsx
```diff
@@ -1,35 +1,39 @@
...
 
 const maxNumberOfEmails = 20;
 type EmailStatuses = Record<InvitationResult, string[]>;
 
+const parseEmails = (text: string): string[] => {
+  return text.split(/[,;\s]/).filter((email) => !isEmpty(email));
+};
+
```
bbenoist, line 36: @nicOwlas I've moved the corresponding code in a `parseEmails` function which could be eventually unit-tested in the future (so that we could cover each case and new bugs related with it).

bbenoist (approved):
@nicOwlas As there was a linter error in the CI, I've took some time to fix it and factorize the code in a dedicated `parseEmaiils` function which:
* reduces the overall complexity of the component (it ain't much but it's a beginning)
* could be reused, extracted or unit-tested in the future.

As I've made the latest changes, I've approved the review so that you can merge it by yourself if you're OK with the changes and if the CI goes green 🚀

typescript/web/src/components/members/new-member-modal.tsx
nicOwlas: Thx for the review and the changes. Let's expedite those small PR and focus on the more important features. Understood for the `Organize imports`. I've enabled it but I cannot find the option to avoid putting grouped imports on the same line.

PR #735, opened 2022-01-04
Update members list max width

### LINKED ISSUE #725: Members list should have a max width
## Problem

When working with a large screen, the members list occupy the whole width which is not great visually.

## Solution

We should use a max width as we did for the classes list. Set it to 5xl (should be the same as in the class list)

## Additional context
Capture d’écran 2022-01-03 à 17 59 55

Jordanlelay:
## Work performed

- Set a max width for the members list
- Update the local workspace alert

## Caveats

Due to the now limited width of the alert, the message is now displayed over 2 lines. @nicOwlas please review the updated UI 

## Resolved issues

Closes #725

### DIFF
--- a/typescript/web/src/components/members/table-actions.tsx
+++ b/typescript/web/src/components/members/table-actions.tsx
@@ -13,6 +13,8 @@ import {
   AlertIcon,
   AlertTitle,
   useColorModeValue as mode,
+  Box,
+  AlertDescription,
 } from "@chakra-ui/react";
 import { useState } from "react";
 import { IoSearch } from "react-icons/io5";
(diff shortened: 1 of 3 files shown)


nicOwlas:
@Jordanlelay could you check why there aren't any deployed version? Thx!

Jordanlelay:
@nicOwlas Not sure why it didn't work in the first place, it's up now

nicOwlas:
Thx! Perfect for the width and the new message on the member tab (local workspace) 👍

valerian (commented):
Looks good to me, waiting for @nicOwlas approval for the design changes

valerian (approved):

typescript/web/src/components/members/table-actions.tsx
```diff
@@ -25,6 +27,19 @@ import { InviteMember } from "./types";
...
+    <Box flex={1}>
+      <AlertTitle>This workspace is private.</AlertTitle>
+      <AlertDescription>
+        Images are stored locally on your device. To collaborate and invite
+        people, switch to an online workspace.
+      </AlertDescription>
+    </Box>
+  </Alert>
+);
+
```
bbenoist, line 42: @Jordanlelay ⚠️ This is not a function ⚠️ Since the PR is merged and nothing broke (nothing broke?), let's keep it for now. But it might be cleaner to keep using function components in the future.

Jordanlelay: You're right, I just kept the things as they were.

It works because it is returned "as is", meaning that `tableActionsForLocalWorkspace` (as well as `tableActionsForOnlineWorkspace`) is of type JSX.Element which is returned by the component `TableActions`.
I will convert them to functional components to keep coherence with the rest of the code

PR #743, opened 2022-01-06
Add sitemap

### LINKED ISSUE #691: Add Sitemap to LabelFlow.ai
In order to improve SEO (to help Google bot crawling and hence ranking our URL), it is important to have xml file on the website. Important as well to help tracking performance. 
Here is an implementation example 

https://blog.logrocket.com/build-sitemap-generator-nextjs/

Jordanlelay:
## Work performed

- Add next-sitemap package

## Results

When using `yarn build` command, `next-sitemap` will generate a sitemap in `typescript/web/public/sitemap.xml`. This file is not commited in the repo, you can generate it using `yarn sitemap` in `typescript/web` folder.

## Resolved issues

Closes #691


typescript/web/.gitignore
```diff
@@ -1,3 +1,4 @@
 
 # Sentry
 .sentryclirc
+public/sitemap.xml
```
bbenoist, line 4:
```suggestion
public/sitemap.xml

```

typescript/web/next-sitemap.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
+  siteUrl: "https://labelflow.ai",
+};
```
bbenoist, line 3: This file shouldn't be required:
[image](https://www.npmjs.com/package/next-sitemap)

Also siteUrl can't be hardcoded to `https://labelflow.ai` it it can change depending on the content of the .env files and environment variables.

bbenoist (changes requested):

typescript/web/next-sitemap.js
Jordanlelay: It is required, it's written in the screenshot you attached.
If you don't put one you get the following error:
```
(node:69018) UnhandledPromiseRejectionWarning: Error: /Users/jordan/dev/labelflow/typescript/web/next-sitemap.js does not exist.
    at Object.exports.getConfigFilePath (/Users/jordan/dev/labelflow/node_modules/next-sitemap/dist/cjs/path/index.js:45:15)
    at /Users/jordan/dev/labelflow/node_modules/next-sitemap/dist/cjs/cli.js:22:35
    at Generator.next (<anonymous>)
    at /Users/jordan/dev/labelflow/node_modules/next-sitemap/dist/cjs/cli.js:8:71
    at new Promise (<anonymous>)
    at __awaiter (/Users/jordan/dev/labelflow/node_modules/next-sitemap/dist/cjs/cli.js:4:12)
    at /Users/jordan/dev/labelflow/node_modules/next-sitemap/dist/cjs/cli.js:20:8
    at Object.<anonymous> (/Users/jordan/dev/labelflow/node_modules/next-sitemap/dist/cjs/cli.js:50:4)
    at Module._compile (internal/modules/cjs/loader.js:1085:14)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1114:10)
(Use `node --trace-warnings ...` to show where the warning was created)
(node:69018) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
(node:69018) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

Regarding the siteUrl, do we already have an env variable for this ?

bbenoist: Yes, I've read the doc but the message below would let us think that it's not necessary anymore 😉 

Actually, we should avoid relying on the public URL at build-time. This is something which is preferrable to allow being changed as close to runtime as possible.

bbenoist: Is the URL really needed?

Jordanlelay: > Actually, we should avoid relying on the public URL at build-time.

But `next-sitemap` is based on the output of `next-build` so why would we want to keep this flexibility ?

> Is the URL really needed?

Yes. If you don't fill it in, you will end up with this kind of output:
```xml
<url><loc>undefined/about</loc><changefreq>daily</changefreq><priority>0.7</priority><lastmod>2022-01-06T16:37:30.326Z</lastmod></url>
```

Where `undefined` is (should be) your siteUrl

bbenoist: > keep this flexibility ?

It's not flexible if the URL is needed during the build process since, for example, we would end-up needing to create 1 docker image per-URL we'd like to use.

If we really can't do things at runtime (e.g. by calling an hypothetical next-sitemap lib which would generate the XML file with SSR), then the call to `next-sitemap` should be moved from the build phase to the deployment phase.

Jordanlelay: I don't see the relation with Docker at this point 🤔

bbenoist: Because, this would either means that customers would have our own URL in their sitemap or that we would be forced use to build a Docker image for each customer once #615.
This is the same problem I had with the IOG URL in https://github.com/labelflow/labelflow/pull/615#issuecomment-996526596

typescript/web/next-sitemap.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
+  siteUrl: "https://labelflow.ai",
```
bbenoist, line 2: Can we at least load the from the .env or something at `build:vercel` time so that:
* it would have a correct value on our preview deployments too
* other users could set their own URL

bbenoist: Maybe that there is a Next function which you can call for that? 🤷

typescript/web/next-sitemap.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
+  siteUrl: process.env.PUBLIC_SITE_URL,
+};
```
bbenoist, line 3: Although its name is not ideal for on-premise, the `VERCEL_URL` variable is already used in our code.

```suggestion
const { VERCEL_URL } = process.env;

if (!VERCEL_URL) {
  throw new Error("Missing required environment variable VERCEL_URL");
}

const SITE_URL =
  VERCEL_URL.includes("://")
    ? VERCEL_URL
    : `https://${VERCEL_URL}`;

module.exports = { siteUrl: SITE_URL };
```

I don't like the optional addition of the `https` prefix at the beginning of `VERCEL_URL` but [that's the way it's expected to be handled](https://github.com/labelflow/labelflow/blob/4b37dbc44ebcd22d103db4ce6f887ee9e4df7cbc/typescript/web/src/constants.ts#L3) for now since it's defined with the prefix our `development.env` file for example.
As such let's assume that it's the way to go and it'll be cleaned once we finish the on-premise PR.

bbenoist (changes requested):
With the suggestions I've made, it should be OK with both Vercel and local (when you call `yarn sitemap` by yourself).

typescript/web/next-sitemap.js
Jordanlelay: Fixed.

By the way, I saw the file you linked the other day and it's written `labelflow.com`, do we own this domain ?

bbenoist: Result of `whois labelflow.com` makes me think that we don't own it yet 😕
@nicOwlas is that normal?

nicOwlas: Yes we don't own it yet. Soon ;-)
https://supabase.com/blog/2021/04/02/supabase-dot-com

typescript/web/next-sitemap.js
```diff
@@ -0,0 +1,20 @@
...
+  : `https://${VERCEL_URL}`;
+
+module.exports = {
+  siteUrl: SITE_URL,
+  exclude: [
+    "/settings/profile",
+    "/debug",
+    "/thank-you",
+    "/request-access",
+    "/graphiql",
```
bbenoist, line 18: The exclude didn't work: https://labelflow-ix6ocxwtv-labelflow.vercel.app/sitemap.xml

bbenoist (changes requested):

typescript/web/next-sitemap.js
Jordanlelay: It's not the good preview, look at [the last one](https://labelflow-8am44mudc-labelflow.vercel.app/sitemap.xml)
They are gone 😉

bbenoist: Oh sorry, I've missed the unique ID in the URL 🤦‍♂️ Then I'll approve your pr. Good job 👏

bbenoist (approved):

PR #744, opened 2022-01-06
Update next to version 12.0.7

Jordanlelay:
## Work performed

- Upgrade next version to 12.0.7

## Caveats

As `next` is generating `next-env.d.ts` on every build, a line has disappeared from ours, probably due to a change from `next`. We could add this file in our `.gitignore` if we want to avoid seeing this file move again in the future.

### DIFF
--- a/typescript/web/package.json
+++ b/typescript/web/package.json
@@ -70,7 +70,7 @@
     "mime-types": "2.1.32",
     "mjml": "4.10.3",
     "multer": "1.4.3",
-    "next": "12.0.2",
+    "next": "12.0.7",
     "next-auth": "4.0.6",
     "next-connect": "0.11.0",
     "next-seo": "4.28.1",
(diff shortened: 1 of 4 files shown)


bbenoist:
@Jordanlelay 👍 to remove `next-env.d.ts` from Git since ignoring it [don't seems to be a problem](https://github.com/vercel/next.js/issues/26533)

bbenoist (approved):

typescript/web/package.json
```diff
@@ -70,7 +70,7 @@
     "mime-types": "2.1.32",
     "mjml": "4.10.3",
     "multer": "1.4.3",
-    "next": "12.0.2",
+    "next": "^12.0.7",
```
bbenoist, line 73:
```suggestion
    "next": "12.0.7",
```

bbenoist (changes requested):

bbenoist (approved):

PR #745, opened 2022-01-06
Optimise image pre loading on labelling interface

### LINKED ISSUE #718: Prioritize loading of current image on Labeling view. And cache-first policy for the already preloaded images
## Objective
Following Vincent's feedback [here](https://labelflow.slack.com/archives/C022PCC0H0W/p1640896185421100):
- Next and previous images are loaded in parallel with the current images which increases the time before one can start labeling

## Tasks
- [x] On the labeling view, prioritize the loading of the current image over its neighbors
- [x] When navigating to the next or previous image, do not reload the already loaded images
- [x] Image and graphql cache-first policy to quickly display the already loaded images and their graph data.

## Potential problem
- If image URLs are signed, the browser will not know it already loaded the image (look for sign URL image cache to find potential solution)

## Investigation
- Check if the url can be used more than one time ? If yes put the URL in the appolo cache

etiennedupont:
## Work performed

Allow the browser to cache calls to the download API route. Since it's a 302 (Temporary redirect) status, the caching is disabled by default so I added explicit headers.

## Resolved issues

Closes #718


nicOwlas:
Etienne could you also prioritize the loading of the current image over the previous/next?

etiennedupont:
> Etienne could you also prioritize the loading of the current image over the previous/next?

Yes clearly, I'm on it.

etiennedupont:
It's working now @nicOwlas , I let you check on the deployment

typescript/web/src/hooks/use-image-pre-fetching.ts
```diff
@@ -63,13 +65,19 @@ export const useImagePrefecthing = () => {
...
-    if (nextImageUrl != null) {
+    if (nextImageUrl != null && canPreFetchNext) {
       fetch(nextImageUrl);
     }
-  }, [nextImageUrl]);
+    setCanPreFetchNext(false);
+  }, [nextImageUrl, canPreFetchNext]);
+  return () => {
+    setCanPreFetchPrevious(true);
+    setCanPreFetchNext(true);
```
bbenoist, line 81: Can't you just have a single state? Or at-least avoid triggering 2 re-renders?

etiennedupont: I agree this is not super satisfying but I couldn't find another way of doing it, the issue is simple: you can't manage the preloading of the next and previous image in the same effect as long as the next and previous image URLs are not updated simultaneously - they're in separate queries. I tried to group the queries to obtain them in a single apollo operation but it fails at the borders, so that's what I ended up doing 🤷

bbenoist: Aaah, I see how it works now 😉 
But my remark is a pure React problem. You can store `canFetchNext` and `canFetchPrevious` in a single object with:

```tsx
type CanFetchState = {
  canFetchNext: boolean;
  canFetchPrevious: boolean;
}

const useCanFetch = () => useReducer<CanFetchState>(
  (oldState, newState) => ({ ...oldState, ...newState }),
  { canFetchNext: false, canFetchPrevious: false }
);

// ...

const [{ canFetchNext, canFetchPrevious }, setCanFetch] = useCanFetch();
```

bbenoist: As `setCanFetch` would accept a partial state as argument, you could then set `canFetchNext`, `canFetchPrevious` or both without triggering 2 state changes 🚀

etiennedupont: Oh, I see, thanks! 
Just one question about the code above, wouldn't it be the same in terms of re renderings to use `useState< CanFetchState >`  over `useReducer` ?

bbenoist: It should but it would force you to always redefine both values at once, even if you only want to set a single one. It's not that important in our case but it makes the code cleaner in general.

etiennedupont: Ok, I'll apply your proposition 👍

bbenoist (approved):

PR #746, opened 2022-01-06
Fix bad aggregates count for workspace > datasets queries

valerian:
## Work performed

- Made the `datasets` queries within a workspace to properly propagate their type, so that aggregate queries can properly know for which parent they are aggregating.

### DIFF
--- a/typescript/db/src/resolvers/workspace.ts
+++ b/typescript/db/src/resolvers/workspace.ts
@@ -153,10 +153,15 @@ const memberships = async (parent: DbWorkspaceWithType) => {
 
 const datasets = async (parent: DbWorkspaceWithType) => {
   const db = await getPrismaClient();
-  return await db.dataset.findMany({
+  const queryResult = await db.dataset.findMany({
     where: { workspaceSlug: parent.slug },
     orderBy: { createdAt: Prisma.SortOrder.asc },
   });
+
+  return queryResult.map((datasetWithoutTypename) => ({
+    ...datasetWithoutTypename,
+    __typename: "Dataset",
+  }));
 };
 
 const stripeCustomerPortalUrl = async (
--- a/typescript/db/src/resolvers/workspace.ts
+++ b/typescript/db/src/resolvers/workspace.ts
@@ -3,6 +3,8 @@ import {
   DbWorkspace,
   DbWorkspaceWithType,
   Repository,
+  addTypename,
+  addTypenames,
 } from "@labelflow/common-resolvers";
 import {
   Membership,
@@ -21,12 +23,6 @@ import { AuthorizationError } from "../repository/authorization-error";
 import { castObjectNullsToUndefined } from "../repository/utils";
 import { stripe } from "../utils";
 
-const addTypename = <TData extends DbWorkspaceWithType | DbWorkspace>(
-  data: TData
-): TData & { __typename: "Workspace" } => {
-  return { ...data, __typename: "Workspace" };
-};
-
 function foundWorkspace<TData extends DbWorkspaceWithType | DbWorkspace>(
   data: TData | null | undefined,
   input: WorkspaceWhereUniqueInput
@@ -44,7 +40,7 @@ const getWorkspace = async (
 ): Promise<DbWorkspaceWithType & { __typename: "Workspace" }> => {
   const data = await repository.workspace.get(where, user);
   foundWorkspace(data, where);
-  return addTypename(data);
+  return addTypename(data, "Workspace");
 };
 
 const workspace = async (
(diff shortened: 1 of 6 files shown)


typescript/db/src/resolvers/workspace.ts
```diff
@@ -153,10 +153,15 @@ const memberships = async (parent: DbWorkspaceWithType) => {
...
   const db = await getPrismaClient();
-  return await db.dataset.findMany({
+  const queryResult = await db.dataset.findMany({
     where: { workspaceSlug: parent.slug },
     orderBy: { createdAt: Prisma.SortOrder.asc },
   });
+
+  return queryResult.map((datasetWithoutTypename) => ({
+    ...datasetWithoutTypename,
+    __typename: "Dataset",
```
bbenoist, line 163: On the long-term, we'll really need to find a way to get rid of these things 😢

In the meantime, can you use an hypothetical `addTypenames` function which would loop over an array of objects to call another hypothetical `addTypename` function?

typescript/web/src/connectors/resolvers/workspace.ts
```diff
@@ -19,8 +23,10 @@ const workspaces = async (
   { repository }: Context
 ) => await repository.workspace.list(args.where);
 
-const datasets = (_parent: any, _args: any, { repository }: Context) => {
-  return repository.dataset.list();
+const datasets = async (_parent: any, _args: any, { repository }: Context) => {
+  const queryResult = await repository.dataset.list();
+
+  return addTypenameToDataArray(queryResult, "Dataset");
```
bbenoist, line 29:
```suggestion
  return addTypenameToDataArray(queryResult, "Dataset");
```

typescript/web/src/connectors/resolvers/workspace.ts
```diff
@@ -19,8 +23,10 @@ const workspaces = async (
   { repository }: Context
 ) => await repository.workspace.list(args.where);
 
-const datasets = (_parent: any, _args: any, { repository }: Context) => {
-  return repository.dataset.list();
+const datasets = async (_parent: any, _args: any, { repository }: Context) => {
```
bbenoist, line 26: I known you're not the author but It's cleaner to use `never` instead of `any` if you don't use these args.

```suggestion
const datasets = async (_parent: never, _args: never, { repository }: Context) => {
```

typescript/db/src/resolvers/workspace.ts
```diff
@@ -153,10 +149,12 @@ const memberships = async (parent: DbWorkspaceWithType) => {
 
 const datasets = async (parent: DbWorkspaceWithType) => {
   const db = await getPrismaClient();
-  return await db.dataset.findMany({
+  const queryResult = await db.dataset.findMany({
     where: { workspaceSlug: parent.slug },
     orderBy: { createdAt: Prisma.SortOrder.asc },
   });
+
```
bbenoist, line 156:
```suggestion
```

typescript/db/src/resolvers/workspace.ts
```diff
@@ -136,7 +132,7 @@ const deleteWorkspace = async (
   const db = await getPrismaClient();
   const data = await db.workspace.findUnique({ where: { id } });
   foundWorkspace(data, args.where);
-  return addTypename(data);
+  return addTypenameToData(data, "Workspace");
```
bbenoist, line 135: Why didn't you kept the `addTypename` function name which was fine IMO?

typescript/common-resolvers/src/utils/add-typename-to-data.ts
```diff
@@ -0,0 +1,21 @@
+export const addTypenameToData = <
+  TData extends Record<string, any>,
+  TTypename extends string
```
bbenoist, line 3: Please don't use `any` in the Record values but either an `object` instead of the Record, a `Record<string, unknown>` or a less generic type (_code suggestion is partial_):

```suggestion
  TRecord extends object,
  TTypename extends string
```

bbenoist (changes requested):

typescript/db/src/resolvers/workspace.ts
valerian: because it felt easier to name `addTypenameToDataArray` along with `addTypenameToData`, with `addTypename` I wasn't sure how to clearly name the array version, but I'm open to suggestions.

bbenoist: Why not `addTypenames`?

valerian: Too easy to confuse and miss the s when re-reading the code, and typing was strong enough to make it fail.
However with your subjected changes on typing, it does properly prevent from misusing the functions, so I will accept your suggestion to revert to `addTypename(s)`

bbenoist (approved):
Seems good to me! Good work 👍

PR #755, opened 2022-01-07
Switch /website footer layout orientation between horizontal on desktop and vertical on mobile

### LINKED ISSUE #701: Fix width of website footer
## Describe the bug

On Mobile the footer of labelflow.ai/website is too wide. We can therefore scroll horizontally which is not wanted.

## To Reproduce

1. Go to labelflow.ai/website on mobile
2. Scroll horizontally.

## Expected behavior

No horizontal scrolling on the website. 

## Device information

 

- **Device:** 
- **OS:** 
- **Browser:** 
- **Version:**

Jordanlelay:
## Work performed

- Update footer for prevent overflow on mobile

## Resolved issues

Closes #701

### DIFF
--- a/typescript/web/src/__stories__/hello.tsx
+++ b/typescript/web/src/__stories__/hello.tsx
@@ -1,7 +1,0 @@
-import React from "react";
-
-export default {
-  title: "web/Hello",
-};
-
-export const World = (args: {}) => <div {...args}>Hello world</div>;
--- a/typescript/web/src/components/website/Footer/Footer.tsx
+++ b/typescript/web/src/components/website/Footer/Footer.tsx
@@ -22,18 +22,18 @@ export const Footer = () => (
         spacing={{ base: "10", lg: "28" }}
       >
         <NextLink href="/website">
-          <Box as="a" href="/website" flex="1">
+          <Box as="a" href="/website" mx={{ base: "auto", lg: "0" }}>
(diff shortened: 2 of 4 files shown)


bbenoist:
It's fixed but I think that the break looks strange.

image

IMO, we should only switch the Stack layout between horizontal and vertical depending on whether the display is large-enough or not.

Jordanlelay:
> It's fixed but I think that the break looks strange.
@bbenoist by break you mean that `Product` and `Learn` are on one line and `Legal` on another one ? 
I was thinking on doing only one column at the beginning but as the width increase it becomes rather empty if each block is alone on it's line

bbenoist:
Yes, this is what I meant.
The thing is that columns widths aren't fixed. Which means that with your changes:
* the 2 first columns take half width on a single row
* the third column takes half width on a single row
* the last column takes full width on a single row

And I feel like it's inconsistent.

If mobile was a priority, I'd say that it needs developable menus but as it's not I've proposed to simply switch between an horizontal and a vertical layout.

Jordanlelay:
Just so you know, 3 `Stack` components can change direction in the footer
I agree with your point, I will update for having one block per row on mobile

bbenoist (approved):
Thanks for your changes. And it looks cleaner centered!

Issue #771, opened 2022-01-11
🌶️  Mock Apollo Client in Jest tests for web


nicOwlas:
## Description

From @bbenoist:
You can see an improvable example in the [CreateWorkspaceModal fixtures](https://github.com/labelflow/labelflow/blob/b52fae9ef1b05e78731eedbe7099168a79e187b9/typescript/web/src/components/workspace-name-input/workspace-name-input.fixtures.tsx#L24) I've recently made.

## Tasks
- [x] Base setup of Apollo mocked provider
- [x] Rewrite Jest tests and Storybook stories
  - [x] Datasets
  - [x] Label Classes
  - [x] Export modal -> #836 
  - [x] Labeling tools -> #842
  - [x] Workspace switcher / create workspace
  - [x] Import button/modal -> #838
  - [x] Hooks -> #841 
  - [ ] Undo store -> #867
- [ ] General testing and cleanup
- [ ] Make sure that everything compiles and tests are ok

(Draft) List of files that should most probably become obsolete and removed in the cleanup phase:
- `web/src/utils/setup-local-db-tests.ts`867

bbenoist:
Closed by #843

Issue #741, opened 2022-01-06
Drop Service Worker


nicOwlas:
## Objective

The service worker is cool to have an app not relying on an online backend and keep the images locally stored. Nevertheless, it adds a layer of complexity which slows the development of LabelFlow's main features.

The goal of this issue is to remove the Service Worker while still keeping a smooth onboarding experience. For the image privacy considerations, the on-premise version of LabelFlow will fill the gap.

### User story - No login required
- User clicks on try it now
- He visualizes the tutorial in his `Personal-[uniqueId]` workspace
- Similarly to the local workspace:
  - Settings are disabled
  - Member page does not allow him to invite people

### Technical questions/difficulties
- We need to create a user & membership in the DB with no prior
- How to display the same personal workspace to the same user everytime. JWT will do the job so long it is not cleared right?
- Need to store a sessionID

### User story - Simplicity
- User clicks on try it now
- Workspace creation modal appears
- Login modal appears
- Tutorial dataset is shown in his newly created workspace
- Workspace is initiated with the community plan
- No need for a specific workspace

## Tasks

- [ ] Remove the service worker code
- [ ] Warn users that their local workspace will be replaced by a personal workplace.
- [ ] Have an home page to redirect to it when one deletes a workspace
  - Excalidraw redirects to the user page
  - We should redirect to another workspace the user is member of
  - If no other workspace, show an empty state with a call to action to create a workspace

## It will allow
- Import: Optimize thumbnail generation
- Optimize DB: focus on Prisma and do not consider Dexie
- Clean typing with specificities due to Dexie
- iog can be used only if the user is signed-in
- Simplify architecture

## Mockup

Workspace empty state page ([link here](https://www.figma.com/file/4J9xyfF5v4B98MuKVO0Y8V/LabelFlow-mockups?node-id=4131%3A14296))
![image](https://user-images.githubusercontent.com/16015322/148402446-50e51f0e-bcb0-4ed0-bc9b-4f624ff9886b.png)

bbenoist:
For the fun (we all have some dirty passions 😜), I've spent an hour removing most of the files specific to the SW (120 files removed, added or modified as of now 🔥).

You can see the result in #748

As of now, `yarn dev:web` is running and I've got most of the app working. I've also built a _quick & very dirty_ `/` homepage which lists workspaces and have a "New workspace" button so that we can start working in parallel with the removal 🚀 

The biggest challenges I see so far before we can merge the changes are (by order of priority):

[...]

Issue #724, opened 2022-01-03
Generate TypeScript types for client-side queries and mutations


bbenoist:
## Is your feature request related to a problem? Please describe

Apollo Client GraphQL (i.e. from the React part of the application) queries and mutations are currently not typed.
This makes the code more subject to errors and hardens the maintenance.

## Describe the solution you'd like (import visuals when possible)

Generate the corresponding types types from all of the `gql` blocks we declare in the code with something like:

```text
npx apollo codegen:generate --localSchemaFile=data/__generated__/schema.graphql --target=typescript --tagName=gql --includes="typescript/web/src/**/*.tsx"
```

## Describe the alternatives you've considered

Manually type everything :-X

## Additional context

We'll also have to use unique names for queries and mutations and remove duplicates since the command currently fails with:

```text
$ npx apollo codegen:generate --localSchemaFile=data/__generated__/schema.graphql --target=typescript --tagName=gql --includes="typescript/web/src/**/*.tsx"                                
CLIError: Error in "Loading queries for Unnamed Project": Error: ️️There are multiple definitions for the `createDataset` operation. Please rename or remove all operations with the duplicated name before continuing.
    at Object.error (/Users/bbenoist/.npm/_npx/68440/lib/node_modules/apollo/node_modules/@oclif/errors/lib/index.js:26:15)
    at Generate.error (/Users/bbenoist/.npm/_npx/68440/lib/node_modules/apollo/node_modules/@oclif/command/lib/command.js:60:23)
    at OclifLoadingHandler.showError (/Users/bbenoist/.npm/_npx/68440/lib/node_modules/apollo/lib/OclifLoadingHandler.js:29:22)
    at OclifLoadingHandler.handle (/Users/bbenoist/.npm/_npx/68440/lib/node_modules/apollo/lib/OclifLoadingHandler.js:14:18)
    at async GraphQLClientProject.scanAllIncludedFiles (/Users/bbenoist/.npm/_npx/68440/lib/node_modules/apollo/node_modules/apollo-language-server/lib/project/base.js:87:9)
    at async Promise.all (index 0) {
  oclif: { exit: 2 },
  code: undefined
}
  ✖ Loading Apollo Project
    → Error initializing Apollo GraphQL project "Unnamed Project": Error: Error in "Loading queries for Unnamed Project": Error: ️️There are multiple defini
…
    Generating query files
[...]

PR #783, opened 2022-01-12
Enhancement/#737 paginate datasets

Jordanlelay:


typescript/web/src/components/datasets/__tests__/upsert-dataset-modal.tsx
```diff
@@ -68,6 +69,11 @@ test("should enable start button when dataset name is not empty", async () => {
 
 test("should create a dataset when the form is submitted", async () => {
   const onClose = jest.fn();
+  // Used to put the query in apollo's cache
+  await client.query({
+    query: getDatasetsQuery,
+    variables: { where: { workspaceSlug: "local" } },
```
bbenoist, line 75: Don't forget that we're supposed be able to do this in less than 2 weeks!

@valerian can we do a meeting tomorrow morning where you'd present us what you're doing on #771 (_this would also let us help you getting everything working faster_)?

typescript/web/src/components/datasets/dataset-list.tsx
```diff
@@ -0,0 +1,121 @@
...
+} from "./dataset-list.context";
+
+export const getDatasetsQuery = gql`
+  query getDatasets($where: DatasetWhereInput) {
+    datasets(where: $where) {
+      id
+    }
+  }
+`;
+const LoadingCard = () => (
```
bbenoist, line 25:
```suggestion
`;

const LoadingCard = () => (
```

typescript/web/src/components/datasets/dataset-list.tsx
```diff
@@ -0,0 +1,121 @@
...
+            datasetName={name}
+            imagesCount={imagesAggregates.totalCount}
+            labelClassesCount={labelClassesAggregates.totalCount}
+            labelsCount={labelsAggregates.totalCount}
+            editDataset={() => {
+              setEditDatasetId(id, "replaceIn");
+            }}
+            deleteDataset={() => {
+              setDeleteDatasetId(id, "replaceIn");
+            }}
```
bbenoist, line 74:
```suggestion
            editDataset={() => setEditDatasetId(id, "replaceIn")}
            deleteDataset={() => setDeleteDatasetId(id, "replaceIn")}
```

typescript/web/src/components/datasets/dataset-list.context.tsx
```diff
@@ -0,0 +1,108 @@
...
+  datasets: Pick<
+    DatasetType,
+    | "id"
+    | "name"
+    | "slug"
+    | "images"
+    | "imagesAggregates"
+    | "labelClassesAggregates"
+    | "labelsAggregates"
+  >[];
```
bbenoist, line 58: Ouch! Thankfully you're working on #724 😅

typescript/web/src/components/datasets/dataset-list.context.tsx
```diff
@@ -0,0 +1,108 @@
...
+      labelClassesAggregates {
+        totalCount
+      }
+    }
+  }
+`;
+
+type UseQueryParamSetter = (
+  value: string,
+  updateType?: "replace" | "push" | "replaceIn" | "pushIn" | undefined
```
bbenoist, line 37: Isn't it an existing type in the library?

Maybe that some of the utility types listed in this doc could help you: https://www.typescriptlang.org/docs/handbook/utility-types.html

typescript/web/src/components/datasets/dataset-list.context.tsx
```diff
@@ -0,0 +1,108 @@
...
+) => void;
+
+export type DatasetListProps = {
+  workspaceSlug?: string | string[];
+  setEditDatasetId: UseQueryParamSetter;
+  setDeleteDatasetId: UseQueryParamSetter;
+};
+export type DatasetListProviderProps = PropsWithChildren<DatasetListProps>;
```
bbenoist, line 45:
```suggestion
};

export type DatasetListProviderProps = PropsWithChildren<DatasetListProps>;
```

typescript/web/src/components/datasets/dataset-list.context.tsx
```diff
@@ -0,0 +1,108 @@
...
+    },
+    skip: workspaceSlug == null || itemCount === 0,
+  });
+  const state = {
+    datasets: datasetsResult?.datasets ?? [],
+    loading: isNil(itemCount) || datasetQueryLoading,
+    workspaceSlug,
+    setEditDatasetId,
+    setDeleteDatasetId,
+  };
```
bbenoist, line 98: Recently, I've started to move such parts in a use `useProviderState` function which returns the state type...
I found that it cleans the code a little bit more as you can write something like:

```tsx
const DatasetListProvider = (props: DatasetListProviderProps) => (
  <DatasetListContext.Provider value={useProviderState(props)}>
    {children}
  </DatasetListContext.Provider>
)
```

typescript/web/src/components/datasets/dataset-list.context.tsx
```diff
@@ -0,0 +1,108 @@
...
+      | "labelClassesAggregates"
+      | "labelsAggregates"
+    >[];
+  }>(getPaginatedDatasetsQuery, {
+    variables: {
+      where: { workspaceSlug },
+      first: perPage,
+      skip: (page - 1) * perPage,
+    },
+    skip: workspaceSlug == null || itemCount === 0,
```
bbenoist, line 90:
```suggestion
    skip: isEmpty(workspaceSlug) || itemCount === 0,
```

_Needs an import from `lodash/fp`_

typescript/web/src/components/datasets/dataset-list.context.tsx
```diff
@@ -0,0 +1,108 @@
...
+  return (
+    <DatasetListContext.Provider value={state}>
+      {children}
+    </DatasetListContext.Provider>
+  );
+};
+
+export const useDatasetList = () => {
+  return useContext(DatasetListContext);
+};
```
bbenoist, line 108:
```suggestion
export const useDatasetList = () => useContext(DatasetListContext);
```

typescript/web/src/components/datasets/dataset-list.tsx
```diff
@@ -0,0 +1,121 @@
...
+    </>
+  );
+};
+
+const PaginationFooter = () => {
+  const { total } = usePagination();
+  return (
+    <>{total > 0 && <PaginationToolbar pos="fixed" bottom={0} w="100%" />}</>
+  );
+};
```
bbenoist, line 87: Can't we share come code with the image pagination (the styling props are the same)?
i.e. make a small `PaginationFooter` component in the `components/pagination` dir

typescript/web/src/components/datasets/dataset-list.tsx
```diff
@@ -0,0 +1,121 @@
...
+      | "name"
+      | "slug"
+      | "images"
+      | "imagesAggregates"
+      | "labelClassesAggregates"
+      | "labelsAggregates"
+    >[];
+  }>(getDatasetsQuery, {
+    variables: { where: { workspaceSlug } },
+    skip: workspaceSlug == null,
```
bbenoist, line 104:
```suggestion
    skip: isEmpty(workspaceSlug),
```

typescript/web/src/components/datasets/dataset-list.tsx
```diff
@@ -0,0 +1,121 @@
...
+const PaginationFooter = () => {
+  const { total } = usePagination();
+  return (
+    <>{total > 0 && <PaginationToolbar pos="fixed" bottom={0} w="100%" />}</>
+  );
+};
+
+export const DatasetList = (props: DatasetListProps) => {
+  const { workspaceSlug } = props;
+  const { data: datasetsResult, loading: totalCountLoading } = useQuery<{
```
bbenoist, line 91: Shouldn't it be in the DatasetListProvider (which could include the `PaginationProvider`)?

typescript/web/src/components/pagination/pagination.context.tsx
```diff
@@ -90,7 +90,8 @@ const usePaginationState = ({
     itemCount,
     total,
     setPagination,
-    setPerPage: (value: number) => setPagination({ perPage: value }),
+    setPerPage: (value: number) =>
+      setPagination({ perPage: value, page: perPage !== value ? 1 : page }),
```
bbenoist, line 94: 👍

typescript/web/src/connectors/repository/utils/list.ts
```diff
@@ -38,14 +39,18 @@ export const list =
       }
 
       const query = table.where(where);
-      const listElements = await query.sortBy(criterion);
+      const listElements = reversed
+        ? await query.reverse().sortBy(criterion)
+        : await query.sortBy(criterion);
```
bbenoist, line 44: What is this for?
Can please you add a comment so that this changed block and the following ont (from l51 to 53) won't look obsure to others?

typescript/web/src/components/datasets/upsert-dataset-modal.tsx
```diff
@@ -102,8 +121,50 @@ export const UpsertDatasetModal = ({
...
+          const updatedDatasets = datasets.concat(createDataset);
+          cache.writeQuery({
+            query: getDatasetsQuery,
+            variables: { where: { workspaceSlug } },
+            data: { datasets: updatedDatasets },
+          });
+        } else {
+          throw new Error("Received null data during dataset creation");
+        }
+      },
```
bbenoist, line 145: In a normal time, I'd have asked to move this block out of the component.

But, since:
1. You're not the author of this big block.
2. We're focusing on deleting #741 

We'll stay with this until we have more time to split the UpsertDatasetModal component into smaller parts 🚀

typescript/web/src/components/datasets/upsert-dataset-modal.tsx
```diff
@@ -102,8 +121,50 @@ export const UpsertDatasetModal = ({
...
+      refetchQueries: ["getPaginatedDatasets"],
+      optimisticResponse: {
+        createDataset: {
+          id: "",
```
bbenoist, line 149: No ID? Won't this create bugs?

typescript/web/src/components/datasets/upsert-dataset-modal.tsx
```diff
@@ -102,8 +121,50 @@ export const UpsertDatasetModal = ({
...
+            query: getDatasetsQuery,
+            variables: { where: { workspaceSlug } },
+            data: { datasets: updatedDatasets },
+          });
+        } else {
+          throw new Error("Received null data during dataset creation");
+        }
+      },
+      refetchQueries: ["getPaginatedDatasets"],
+      optimisticResponse: {
```
bbenoist, line 147: Can we really be optimistic with Dataset creation (as we have to upload files right after)?

bbenoist (changes requested):
I've made quite a lot of comments but it's a very good work 👏

typescript/web/src/components/datasets/dataset-list.context.tsx
Jordanlelay: No it's not an existing type but I can use a exported one for the different strings

typescript/web/src/components/datasets/dataset-list.tsx
Jordanlelay: This query is used before mounting `DatasetListProvider`. The return is passed to `PaginationProvider` and used then for determine if more data has to be loaded.

As `DatasetListProvider` depends on data from `PaginationProvider`, no it shouldn't be the way you ask in my opinion 😉

typescript/web/src/components/datasets/dataset-list.tsx
Jordanlelay: Fixed

typescript/web/src/components/datasets/dataset-list.tsx
Jordanlelay: Fixed

typescript/web/src/components/datasets/dataset-list.context.tsx
Jordanlelay: Fixed

typescript/web/src/components/datasets/dataset-list.context.tsx
Jordanlelay: Complexity was low enough, but I did it anyway. Not sure I will use it everytime

typescript/web/src/components/datasets/dataset-list.context.tsx
Jordanlelay: Fixed

typescript/web/src/components/datasets/dataset-list.context.tsx
Jordanlelay: Fixed

typescript/web/src/components/datasets/dataset-list.tsx
Jordanlelay: Fixed

typescript/web/src/components/datasets/dataset-list.tsx
Jordanlelay: Fixed

typescript/web/src/connectors/repository/utils/list.ts
Jordanlelay: Fixed

typescript/web/src/components/datasets/upsert-dataset-modal.tsx
Jordanlelay: Fixed

Jordanlelay: No as the refetchQuery `getPaginatedDatasets` will override the missing ID

typescript/web/src/components/datasets/upsert-dataset-modal.tsx
Jordanlelay: I'm not sure I understand the question

typescript/web/src/components/datasets/__tests__/upsert-dataset-modal.tsx
Jordanlelay: Actually this is a small hack to push something consistent in apollo's cache as my components rely on it

bbenoist: OK. Is is possible to use a different workspace name then or is the test still tied to the SW?

typescript/web/src/components/datasets/dataset-list.context.tsx
bbenoist: I've seen you've made to code to use [`UrlUpdateType`](https://github.com/pbeshai/use-query-params/blob/master/packages/use-query-params/src/types.ts#L11) now. That's what I've asked for actually, perfect :)
The sentence about the utility types was just in case the type was only declared as an argument of the query-params setter function (i.e. `setParam(........, updateType: "replace" | "push" | "replaceIn" | "pushIn" | undefined)` which would have forced you to use something like `Parameters<typeof setParam>[1]`.

valerian (changes requested):
It doesn't seem that you handle the possible issues with the paginated query's cache, some being more critical than others.

Keep in mind that by default Apollo caches the return value of a query based on its parameters, so currently have a different cache entry for each combination of `{ where: { datasetId }, first, skip }`.

The second important thing to keep in mind is that refetchQueries will only refetch a given query _with the last parameters it was called with_, meaning that it refetches the current page only, not the other pages, which will still keep using their previous cached values.

This results in those problems:
- If you delete an item on page 1, the refetchQueries will properly display page 1. But since you deleted an item, all indexes should be shifted now, and if pages 2, 3, etc... are in cache, they will have the wrong values, same problem if you add an item. (solution: invalidate all the whole paginated cache, not just the last query, when adding/removing items)
- (Minor issue) Since each parameters combo has a different cache entry, pages with different "per page" options will not be able to reuse the cached data from other queries. (solution: implement a custom merging cache like I did)

valerian (approved):

typescript/web/src/components/datasets/__tests__/upsert-dataset-modal.tsx
Jordanlelay: Yes it's possible, I will use a different name

bbenoist (approved):

PR #790, opened 2022-01-13
Fix exports not working from datasets in workspaces

### LINKED ISSUE #751: Make dataset export work in online workspaces
## Is your feature request related to a problem? Please describe

Robustify export, it fails to export labels only in some cases

## Describe the solution you'd like (import visuals when possible)

- [ ] Export generates an .xml file if one image is not found ([example this one](https://labelflow.ai/demo/datasets/tutorial-2/images))
- [ ] For some datasets, example [this one](https://labelflow.ai/omexom-arabie-saoudite/datasets/2018-12-19-dammam-tower-557-type-da_tower-556-type-da-tower-557-type-da-cjptzhu77h6h80790sna6mpjm/images), the export of labels to coco format (not exporting the images, also fails though no images are missing)

## Additional context

Due to remote vs local workspace?

valerian:
## Work performed

- Fix exports from workspace datasets using the wrong URL (used the upload URL instead of the download URL).
- Fix downloaded files of exports having no extension. 

## Results

Exports from workspace datasets should now produce properly downloadable files.

## Resolved issues

Fixes #751

### DIFF
--- a/typescript/common-resolvers/src/export/index.ts
+++ b/typescript/common-resolvers/src/export/index.ts
@@ -41,15 +41,14 @@ const exportDataset = async (
 ) => {
   const fileExport = await generateExportFile(args, { repository, req }, user);
   const origin = getOrigin(req);
-  const outUrl = (
+  const { uploadUrl, downloadUrl } =
     await repository.upload.getUploadTargetHttp(
       // FIXME make this Url disappear at some point...
       uuidv4(),
       origin
-    )
-  )?.uploadUrl;
-  await repository.upload.put(outUrl, fileExport, req);
-  return outUrl;
+    );
+  await repository.upload.put(uploadUrl, fileExport, req);
+  return downloadUrl;
 };
 
 export default {
--- a/typescript/web/src/components/export-button/export-modal/export-dataset.ts
+++ b/typescript/web/src/components/export-button/export-modal/export-dataset.ts
@@ -57,8 +57,10 @@ export const exportDataset = async ({
   const blobDataset = await (await fetch(exportDatasetUrl)).blob();
   const url = window.URL.createObjectURL(blobDataset);
   const element = document.createElement("a");
+  const extension =
+    format === ExportFormat.Yolo || options.coco?.exportImages ? "zip" : "json";
   element.href = url;
-  element.download = datasetName;
+  element.download = `${datasetName}.${extension}`;
   setIsExportRunning(false);
   element.click();
 };


typescript/web/src/components/export-button/export-modal/index.tsx
```diff
@@ -104,8 +104,10 @@ const exportDataset = async ({
   const blobDataset = await (await fetch(exportDatasetUrl)).blob();
   const url = window.URL.createObjectURL(blobDataset);
   const element = document.createElement("a");
+  const extension =
+    format === ExportFormat.Yolo || options.coco?.exportImages ? "zip" : "json";
   element.href = url;
-  element.download = datasetName;
+  element.download = `${datasetName}.${extension}`;
```
bbenoist, line 110: Was it because the extension was missing in the file name?
Have you made sure it didn't broke anything with local? _(this will be useful for #741 since we need export to work)_

typescript/web/src/components/export-button/export-modal/index.tsx
```diff
@@ -104,8 +104,10 @@ const exportDataset = async ({
   const blobDataset = await (await fetch(exportDatasetUrl)).blob();
   const url = window.URL.createObjectURL(blobDataset);
   const element = document.createElement("a");
+  const extension =
+    format === ExportFormat.Yolo || options.coco?.exportImages ? "zip" : "json";
```
bbenoist, line 108: To be more future-proof, I'd have factorized things a little bit. But I doubt it's really useful since me might remove this part once we drop the SW.
Did you have time to understand why it downloaded the file in a blob or asked old devs?

typescript/web/src/components/export-button/export-modal/index.tsx
valerian: Extension was missing in the filename, but in local the browser automatically added the extension based on the mime type of the data, however on workspaces we only ever get a generic type `application/octet-stream`.

typescript/web/src/components/export-button/export-modal/index.tsx
valerian: Yes I was about to refactor all that, but thought that it would be a waste of time since it should probably get re-refactored again soon.

No I didn't ask old devs, but I understand the reason: it's to obfuscate the URL to the download API that creates the signed URL: it would allow getting new signed URLs anytime, since the file is never deleted from our bucket.
See that comment: https://github.com/labelflow/labelflow/issues/751#issuecomment-1012074286

typescript/web/src/components/export-button/export-modal/index.tsx
bbenoist: Ah I see. It's strange that web-browsers have a different behavior depending on whether the file is local of not... Thanks for the explanation.
The solution seems OK then. Nice catch 👍

valerian:
> @valerian I've asked the YOLO and COCO exports to include images but the file got the XML extension:
In which environment did you do this test? Can you give me stpes to reproduce?
I tried exports with SW and online workspace, with and without images, on local dev server or on vercel deployment, all of them work for me.

typescript/web/src/components/export-button/export-modal/index.tsx
valerian: > Ah I see. It's strange that web-browsers have a different behavior depending on whether the file is local of not... Thanks for the explanation. The solution seems OK then. Nice catch 👍

It's not depending on whether the file is local or not, it's depending on whether the file is stored in Dexie or in S3 :)

bbenoist:
@valerian

* Vercel preview deployment (as such, signed in with email)
* Windows 11 with the latest Chromium-based Edge version
* The dataset only had a single image with an iog label on it
* I've asked to export labels

valerian:
@bbenoist 
Still can't reproduce, if it's an online workspace can you give me access to it?
In the meantime since it doesn't seem related to the current pull request we should try to merge it soon.

bbenoist:
@valerian After retrying again, I still can't reproduce the issue. Perhaps it was a sign-in problem or something similar 🤷‍♂️ 
OK to merge it as-is. We'll keep an eye on it so that we'll be sure that it won't reproduce again.

bbenoist (approved):

PR #795, opened 2022-01-13
Fix pagination issues when no items

### LINKED ISSUE #793: paginatedImagesQuery returns a 400 when there is no image in a dataset
When a dataset has no images, we call `paginatedImagesQuery` which returns a 400 error.

valerian:
## Work performed

- Update `usePaginationState` to prevent `page` to be lower than `1`
- Update `ImagesListProvider` to prevent the image list query to be executed whenever the total count of items is `0`
- Some code cleanup such as the removal of unused aggregates queried on the dataset images page.

## Results

- No more 400 error if there is no image in a dataset.
- No more images query on datasets that have no images.
- No more extra database queries for aggregates that won't be used.

## Resolved issues

Resolves #793


typescript/web/src/pages/[workspaceSlug]/datasets/[datasetSlug]/images/index.tsx
```diff
@@ -29,12 +29,6 @@ export const datasetDataQuery = gql`
       imagesAggregates {
         totalCount
       }
-      labelsAggregates {
```
Comment (Jordanlelay) [inline typescript/web/src/pages/[workspaceSlug]/datasets/[datasetSlug]/images/index.tsx:32]: Why are you removing this part ? 🤔

Comment (valerian) [reply typescript/web/src/pages/[workspaceSlug]/datasets/[datasetSlug]/images/index.tsx:32]: because it's never ever used, and makes extra SQL queries

bbenoist (approved):

bbenoist (approved):

PR #748, opened 2022-01-07
Switch the app to online workspaces

bbenoist:
## Work performed

I have:

* deleted most of the code related to the service worker
* created a very basic homepage so that we can continue working on #741 in //

## Results

We should now have a single _GraphQL resolvers_, _DB repositories_ and _file storage_ implementation.

## Problems encountered

See #741 for details

## Caveats

It's a long debate...

## Resolved issues

Closes #741

PR #800, opened 2022-01-17
Generate types for web GraphQL queries and mutations

### LINKED ISSUE #724: Generate TypeScript types for client-side queries and mutations

Jordanlelay:
## Work performed

- De-duplicate every query and mutation using the `gql` function
- Give names to every anonymous query and mutation making use of `gql` ([more info](https://github.com/apollographql/apollo-tooling/issues/184#))
- Add a resolution section in package.json (to fix [this problem](https://github.com/apollographql/apollo-tooling/issues/577), [yarn doc here](https://classic.yarnpkg.com/en/docs/selective-version-resolutions))
- Add a webpack option for generating queries and mutation types on every emit (each time webpack compile, which is maybe overkill, maybe we can find a [better hook here](https://webpack.js.org/api/compiler-hooks/))

## Results

Apollo queries and mutations will have types generated automatically

## Caveats

The type generation won't work if:
 - We use same names for queries / mutations
 - We use anonymous queries / mutations

## Resolved issues

Closes #724


typescript/web/next.config.js
```diff
@@ -72,16 +71,28 @@ module.exports = withSentryConfig(
...
+              compiler.hooks.afterEmit.tap("AfterEmitPlugin", (compilation) => {
+                exec(
+                  'yarn apollo codegen:generate --localSchemaFile=../../data/__generated__/schema.graphql --target=typescript --tagName=gql --includes="src/**/*.tsx"',
+                  (err, stdout, stderr) => {
+                    if (stdout) process.stdout.write(stdout);
+                    if (stderr) process.stderr.write(stderr);
+                  }
+                );
+              });
+            },
```
bbenoist, line 93: As I understand this code, the build wouldn't fail if codegen had an error. Is that right?

Also, can you factorize it into a `generate` function which would simply call `yarn generate`.
This `generate` command would make sure that **_both_** the apollo generator and the GraphQL codegen would be called (so that we could also gitignore the output of GraphQL very soon).

typescript/web/src/components/dataset-classes/upsert-class-modal/create-label-class.mutation.ts
```diff
@@ -12,9 +12,9 @@ import { useCallback } from "react";
 import { v4 as uuid } from "uuid";
 import { DATASET_LABEL_CLASSES_QUERY } from "./dataset-label-classes.query";
 
-export const CREATE_LABEL_CLASS_MUTATION = gql`
+export const createLabelClassMutation = gql`
```
bbenoist, line 15: We agreed to use `CONSTANT_CASE` (to denote that these globals should be considered as _deeply-immutable_). Why did you do the inverse?

Jordanlelay: It's a rebase conflict that I resolved. It's specifically designed to annoy you

bbenoist: Ahah! Of course it's not 😅
I now understand why and I agree that it was easier to resolve the conflict this way 👍 
Sorry for asking such questions while it was still a draft.

typescript/web/next.config.js
Jordanlelay: Do you mean the build resulting of the command `yarn build:web` ?

bbenoist: In the sentence _`the build wouldn't fail if codegen had an error`_ I meant the process which called to this code (whichever it is, dev or build mode).

If your question objective was to make sure that I didn't meant _"please make this function called also during `yarn build:web`"_, then no.
But why not if it's possible and don't complexifies things 🚀  At-least this would be more respectful of the DRY principle...

Jordanlelay: Then yes you are right it would not fail, it just fires an error in console but it's emitted after the build

bbenoist: Why wouldn't you want it to fail?

Jordanlelay: Because we never discuss it 🤷

bbenoist: Because that's a common rule of thumb we do (_nearly_) everywhere.
In a robust process, regardless of the language its been written with, you should fail as soon as an error occur in a non-optional step. Not ignore it and print a message on stderr.
They are cases where you'll not want the error to completely stop your process (_mostly with servers and GUI_) but these will be handled with `catch` clauses placed at the highest level in the stack so that you keep the lower-levels robust and side-effect free.
For example, in our specific case, we dont wan't to kill the high-level watcher which restarts the "build" at each change but we want to prevent the rest of the "build" from continuing if we haven't generated the files.

bbenoist (changes requested):
Thanks for doing this painful work 👏 

Now that the generator is wired, we still have to:

- [x] Resolve the comments I've made in this review
- [x] Live-test the changes on my machine and see together if it's possible to optimize the webpack hook as suggested in your PR description.
- [ ] Rename queries and mutations definitions with CONSTANT_CASE
- [x] Meet with @valerian again to see if he could merge the branch soon so that he'd have less conflicts once merging-back his branch into #748
- [ ] There is also this `utils/shared-queries` directory which I'd like analyse deeper. At a first sight, it looks like something risky. I've not looked at it in details yet but I'd like to see whether some queries are actually just used in duplicate React components which could be merged.
- [ ] In a new PR (_so that we can merge this one earlier and minimize Git conflicts over time_), reuse the generated types in `useQuery`, `useMutation`, `client.query` and `client.mutate` and propagate them to their callers.

package.json
```diff
@@ -88,5 +89,9 @@
     "typescript": "4.3.5",
     "uuid": "8.3.2",
     "webpack": "5.38.1"
+  },
+  "resolutions": {
+    "@types/graphql": "14.5.0",
+    "graphql": "15.5.1"
   }
```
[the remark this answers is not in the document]
Jordanlelay: No it's in package.json ([see here](https://classic.yarnpkg.com/en/docs/selective-version-resolutions))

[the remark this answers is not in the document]
bbenoist: The link doesn't really tells that it's not doable from `.yarnrc.yml`. But I've found the information [here](https://yarnpkg.com/configuration/yarnrc#packageExtensions), you're right, it's only usable in the `package.json` file.

As I now understand better what's the purpose of this option and the difference with `packageExensions`, I think that it could be avoidable by pinning exact versions in the `package.json` dependencies instead of prefixing them with `^` ([doc](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#dependencies)).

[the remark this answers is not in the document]
Jordanlelay: I do have exact versions of `graphql` package. However, `apollo` package comes with a `graphql` dependency which is probably not the same than ours.

So if you remove the `resolutions` block, you will get the following error:

```
Error: Command failed: yarn codegen:all
    Error: Cannot use GraphQLScalarType "ID" from another module or realm.

    Ensure that there is only one instance of "graphql" in the node_modules
    directory. If different versions of "graphql" are the dependencies of
    other
    relied on modules, use "resolutions" to ensure only one version is
    installed.

    https://yarnpkg.com/en/docs/selective-version-resolutions

    Duplicate "graphql" modules cannot be used at the same time since
    different
    versions may have different capabilities and behavior. The data from one
    version used in the function from another could produce confusing and
    spurious results.
```

PS: we can remove the line about graphql types

typescript/web/.gitignore
```diff
@@ -4,3 +4,5 @@
 
 next-env.d.ts
 public/sitemap.xml
+
+**/__generated__/
```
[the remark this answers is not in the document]
Jordanlelay: Fixed

typescript/web/next.config.js
```diff
@@ -72,16 +71,28 @@ module.exports = withSentryConfig(
...
           }),
+          {
+            apply: (compiler) => {
+              compiler.hooks.afterEmit.tap("AfterEmitPlugin", (compilation) => {
+                exec("yarn generate", (err, stdout, stderr) => {
+                  if (stdout) process.stdout.write(stdout);
+                  if (stderr) process.stdout.write(stderr);
+                  if (err) {
+                    process.exit(1);
+                  }
```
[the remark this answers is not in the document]
Jordanlelay: Fixed

typescript/web/package.json
```diff
@@ -11,7 +11,8 @@
     "type-check": "tsc",
     "bundlewatch": "bundlewatch --config ./bundlewatch.config.js",
     "release": "release-it",
-    "release:update": "cd ../release && yarn update"
+    "release:update": "cd ../release && yarn update",
+    "generate:apollo": "yarn apollo codegen:generate --localSchemaFile=../../data/__generated__/schema.graphql --target=typescript --tagName=gql --includes=\"src/**/*.tsx\""
```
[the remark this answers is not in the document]
Jordanlelay: Fixed

typescript/web/src/components/welcome-manager/welcome-modal.tsx
```diff
@@ -26,7 +26,7 @@ type WelcomeModalParam =
   | "closed"; // Force it to be closed and never open
 
 export const getDatasetsQuery = gql`
-  query getDatasets($where: DatasetWhereInput) {
+  query getDatasetsName($where: DatasetWhereInput) {
```
valerian, line 29:
```suggestion
  query getDatasetsNames($where: DatasetWhereInput) {
```

package.json
```diff
@@ -25,14 +25,17 @@
...
     "codegen:ci": "graphql-codegen --config ./codegen.yml && git diff --exit-code",
+    "codegen:data": "graphql-codegen --config ./codegen.yml",
+    "codegen:web": "cd typescript/web && yarn codegen",
     "storybook": "start-storybook -s ./typescript/web/public -p 6006",
     "storybook:all": "start-storybook -s ./typescript/web/public -p 6006",
     "build-storybook": "build-storybook",
     "chromatic": "npx chromatic",
-    "release": "cd ./typescript/web && yarn release"
+    "release": "cd ./typescript/web && yarn release",
+    "generate": "yarn codegen && yarn generate:apollo"
```
bbenoist, line 38: Do we still need this script?

valerian (changes requested):
I see that unfortunately you did a lot of work on __tests__ file, because soon gql queries won't used in web tests anymore

package.json
Jordanlelay: In case someone wants to generate the types without running the web server ?

bbenoist: But `generate:apollo` does not even seems to exist...

`yarn generate` was however called in `.github/workflows/ci.yml but it's always for the `db`.

Maybe we could just rename the `generate` into `codegen` and harmonize things and use a `codegen:db` script and integrate it like what you've done for the `data` and `web` `codegen` scripts?

Jordanlelay: No you were right, this command is not used, we replace it when we harmonize the `codegen` commands earlier, I will delete it

typescript/web/package.json
```diff
@@ -2,16 +2,17 @@
...
+    "build": "cd ../db && yarn codegen:all && cd ../web && next build",
     "build:vercel": "yarn build && yarn sitemap && cd ../db && sh ./deploy-vercel.sh",
     "sitemap": "next-sitemap",
     "start": "next start",
     "type-check": "tsc",
     "bundlewatch": "bundlewatch --config ./bundlewatch.config.js",
     "release": "release-it",
-    "release:update": "cd ../release && yarn update"
+    "release:update": "cd ../release && yarn update",
+    "codegen": "yarn apollo codegen:generate --localSchemaFile=../../data/__generated__/schema.graphql --target=typescript --tagName=gql --includes=\"src/**/*.tsx\""
```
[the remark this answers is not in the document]
Jordanlelay: Nice catch 👍  I copy pasted the command without thinking it

valerian (commented):
Have you checked that the query names you changed did not break any `refetchQueries` strings?

valerian (approved):

Jordanlelay:
> Have you checked that the query names you changed did not break any `refetchQueries` strings?

No I didn't.
However, it should not, as the `refetchQueries` strings across the app make use of the name of the query itself, so renaming `gql` objects won't affect that

bbenoist:
OK guys, it looks like we can now merge this issue into `enhancement/delete-service-worker`! Good job with the code and its review 👏

@Jordanlelay Can you please fix the minor conflict with the package files and make sure that your branch is up-to-date so we can start reusing the types right-now and proceed to the next steps of the SW removal?

@valerian Prepare for hell with #803 :trollface: We can mob the rebase (or merge) if you want :)

valerian:
> @valerian Prepare for hell with #803 :trollface: We can mob the rebase (or merge) if you want :)

I am not afraid, I believe that the conflicts will be rather simple to resolve

Jordanlelay:
Well, unfortunately, @valerian was right, there are problems with some `refetchQueries`. I found one but it might have others so keep an eye on the console (I suspect the problem was there before as it made no sense to refetch that)

bbenoist:
> > @valerian Prepare for hell with #803 :trollface: We can mob the rebase (or merge) if you want :)
>
> I am not afraid, I believe that the conflicts will be rather simple to resolve

Yes it shouldn't be complex. Just big 😉

bbenoist:
> Well, unfortunately, @valerian was right, there are problems with some refetchQueries. I found one but it might have others so keep an eye on the console (I suspect the problem was there before as it made no sense to refetch that)

OK. We'll try to test the app intensively and fix the console issues we find before merging #748 🧹 

In the meantime I've found `MUTATION_CREATE_IMAGE` which should be named `CREATE_IMAGE_MUTATION` to match with the other queries and mutations we have in the code. Can you do the change while you're rebasing/squashing the branch?

typescript/web/src/pages/[workspaceSlug]/datasets/[datasetSlug]/images/[imageId].tsx
```diff
@@ -42,23 +43,14 @@ const LabelingTool = dynamic(
 );
 
 const imageQuery = gql`
-  query image($id: ID!) {
+  query imageName($id: ID!) {
```
Comment (valerian) [inline typescript/web/src/pages/[workspaceSlug]/datasets/[datasetSlug]/images/[imageId].tsx:46]: another refetched query to update (potentially)

typescript/web/src/components/dataset-classes/dataset-classes.query.ts
```diff
@@ -2,7 +2,10 @@ import { gql, useQuery } from "@apollo/client";
 import { DatasetClassesQueryResult } from "./types";
 
 export const DATASET_LABEL_CLASSES_QUERY = gql`
-  query getDatasetLabelClasses($workspaceSlug: String!, $datasetSlug: String!) {
+  query getDatasetLabelClassesWithTotalCount(
```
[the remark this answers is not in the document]
Jordanlelay: That's the one I've spotted

typescript/web/src/components/welcome-manager/welcome-modal.tsx
```diff
@@ -26,7 +26,7 @@ type WelcomeModalParam =
   | "closed"; // Force it to be closed and never open
 
 export const getDatasetsQuery = gql`
-  query getDatasets($where: DatasetWhereInput) {
+  query getDatasetsNames($where: DatasetWhereInput) {
```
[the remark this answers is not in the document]
Jordanlelay: It does not produce warning but I've updated it

typescript/web/src/components/dataset-classes/dataset-classes.query.ts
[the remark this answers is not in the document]
valerian: There used to be two queries with the same name `getDatasetLabelClasses `, and I think both of them made sense to get refetched in `typescript/web/src/components/import-button/import-images-modal/modal-dropzone/import-dropped-files/import-datasets.tsx`, so can you add this one as well?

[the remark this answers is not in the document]
Jordanlelay: I'm not sure I understand, you want me to add `getDatasetLabelClassesWithTotalCount` in the `refetchQueries` of `typescript/web/src/components/import-button/import-images-modal/modal-dropzone/import-dropped-files/import-datasets.tsx` ?

valerian (approved):

PR #803, opened 2022-01-17
Mock apollo client in jest tests for web

valerian:
- [x] Base setup of Apollo mocked provider
- [x] Rewrite Jest tests and Storybook stories
  - [x] Datasets
  - [x] Label Classes
  - [ ] Export modal -> #836 
  - [ ] Labeling tools
  - [ ] Workspace switcher
  - [ ] Import button/modal
  - [ ] Hooks
- [ ] General testing and cleanup


typescript/web/src/utils/testing/mock-apollo.tsx
```diff
@@ -0,0 +1,42 @@
...
+  Resolvers,
+} from "@apollo/client/core";
+import {
+  MockedProvider,
+  MockedProviderProps,
+  MockedResponse,
+} from "@apollo/client/testing";
+import { Mutation, Query } from "@labelflow/graphql-types";
+
+export type ApolloMockedResponse = MockedResponse<
```
bbenoist, line 16:
```suggestion
export type MockedApolloResponse = MockedResponse<
```

typescript/web/src/utils/testing/mock-apollo.tsx
```diff
@@ -0,0 +1,42 @@
...
+  addTypename?: boolean;
+  defaultOptions?: DefaultOptions;
+  cache?: ApolloCache<TSerializedCache>;
+  resolvers?: Resolvers;
+  childProps?: object;
+  children?: any;
+  link?: ApolloLink;
+}
+
+export class ApolloMockedProvider extends MockedProvider {
```
bbenoist, line 32:
```suggestion
export class MockedApolloProvider extends MockedProvider {
```

typescript/web/src/utils/testing/mock-apollo.tsx
```diff
@@ -0,0 +1,42 @@
+import { PropsWithChildren } from "react";
```
bbenoist, line 1: We it's still possible, can we rename the `testing` directory as `tests` so that it's consistent with the `__tests__` directories we have everywhere?

typescript/web/src/utils/testing/mock-data.ts
```diff
@@ -0,0 +1,10 @@
+export const mockWorkspaceSimple = {
```
bbenoist, line 1: * Can we call the file `data.fixtures.ts`?
* Can you use CONSTANT_CASE for deep-readonly values?
* Why didn't you reused the GraphQL types defined in `@labelflow/graphql-types`?

typescript/web/src/utils/setup-local-db-tests.ts
```diff
@@ -4,6 +4,7 @@ import { initMockedDate } from "@labelflow/dev-utils/mockdate";
 import { getDatabase } from "../connectors/database";
 import { client } from "../connectors/apollo-client/schema-client";
 
+// TODO SW probably obsolete and to delete
 export function setupTestsWithLocalDatabase() {
```
bbenoist, line 8: - [ ] We have to fix this TODO in this PR. But I agree it can be delayed to the end

.eslintrc
```diff
@@ -22,12 +22,16 @@
...
           "**/__tests__/**/*.ts",
           "**/__stories__/**/*.ts",
+          "**/storybook/**/*.ts",
           "**/*.test.tsx",
           "**/*.stories.tsx",
+          "**/*.fixtures.tsx",
           "**/__tests__/**/*.tsx",
-          "**/__stories__/**/*.tsx"
+          "**/__stories__/**/*.tsx",
+          "**/storybook/**/*.tsx"
```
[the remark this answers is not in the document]
valerian: Sure I can do that (those changes come from _your_ commit though :P)

typescript/web/src/components/datasets/__tests__/upsert-dataset-modal.tsx
```diff
@@ -1,66 +1,112 @@
...
+        updateDataset: { id: mockDatasetSimple.id },
+      },
+    })),
+  },
+};
 
 const renderModal = (props = {}) => {
   return render(<UpsertDatasetModal isOpen onClose={() => {}} {...props} />, {
-    wrapper: Wrapper,
+    wrapper: getApolloMockWrapper(Object.values(mockQueries)),
```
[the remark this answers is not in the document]
valerian: We use a record in order to more easily access specific members, which might implement `newData` functions monitored by jest `fn`.
It also makes the mock queries intention/meanings easier to understand by naming them appropriately.

I will refactor my implementation of our ApolloMockProvider to accept a record.

typescript/web/src/utils/setup-local-db-tests.ts
valerian: Yes my intention is do cleanup this code and other similar ones in this PR, but only after all tests are running.

typescript/web/src/utils/testing/mock-data.ts
valerian: > * Can you use CONSTANT_CASE for deep-readonly values?
What do you mean?

typescript/web/src/components/datasets/__tests__/delete-dataset-modal.tsx
```diff
@@ -1,99 +1,65 @@
...
+} from "../../../utils/testing/mock-apollo";
+import {
+  DeleteDatasetModal,
+  getDatasetByIdQuery,
+  deleteDatasetByIdMutation,
+} from "../delete-dataset-modal";
+import { mockDatasetSimple } from "../../../utils/testing/mock-data";
 
-setupTestsWithLocalDatabase();
+const mockQueries: Record<string, ApolloMockedResponse> = {
```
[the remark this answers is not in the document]
valerian: I agree that using a "standardised" fixtures file for consistency is good, but this component doesn't use Apollo in its stories, so I didn't make a fixtures file for it, since it would only be used for `__tests__` yet.

What do you think should be done in this kind of case?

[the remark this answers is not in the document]
bbenoist: Some goals of such harmonization:
* Reduce the amount of mental charge needed to understand tests made by others.
* Quickly get new tests up and running. By using "standardized" names, we can just scaffold new tests from existing ones without having to rename all of the variables and functions present in every tests.
* Encourage factorization. Since the blocks are already there, it just becomes easier to reuse them in other contexts.

In that sense, I'd say :
* At-least try to use the same `APOLLO_MOCKS` name in every tests so that we can start to have a reusable patterns.
* I'd be less strict with the `.fixtures` files I think...
  They are mostly useful to share data between tests, yes. Not only a single jest and storybook test couple, but also between multiple tests in the app.
  But besides that, it doesn't adds that much value so I guess we can just keep the correct names in the `.test` and `.stories` files as long as the code blocks are not too big (_which is what I don't really like with the Apollo mocks_) and don't need to be shared. We could still extract them in a `.fixtures` file.

This is the rule of thumb I've ended-up with for now, the `type TestCase`, `const TEST_CASES`, `function runTest` pattern you can see in all of my PR... But it still needs to be refined with feedbacks such as yours :)

I know it can seems a bit overthinking sometimes but at the end, it's a combination of those little things which makes us design new features faster while minimizing the impact on robustness.

typescript/web/src/utils/testing/mock-apollo.tsx
bbenoist: OK for `stories` 👍

typescript/web/src/components/dataset-classes/upsert-class-modal/upsert-class-modal.fixtures.ts
```diff
@@ -0,0 +1,83 @@
...
+    request: {
+      query: CREATE_LABEL_CLASS_MUTATION,
+      variables: MATCH_ANY_PARAMETERS,
+    },
+    result: jest.fn((variables) => ({
+      data: {
+        createLabelClass: {
+          id: variables?.id ?? "DEFAULT-NEW-ID",
+          name: variables?.name ?? "DEFAULT-NEW-NAME",
+          color: variables?.color ?? "DEFAULT-NEW-COLOR",
```
bbenoist, line 45: Can't we generate random values if variables are undefined?

typescript/web/src/components/dataset-classes/upsert-class-modal/upsert-class-modal.fixtures.ts
```diff
@@ -0,0 +1,83 @@
...
+import { LABEL_CLASS_EXISTS_QUERY } from "./label-class-exists.query";
+import { UPDATE_LABEL_CLASS_NAME_MUTATION } from "./update-label-class-name.mutation";
+import {
+  MOCK_DATASET_WITH_CLASSES,
+  MOCK_LABEL_CLASS_SIMPLE,
+} from "../../../utils/tests/data.fixtures";
+
+export const MOCK_LABEL_CLASS_UPDATED_NAME = "My New Class Name";
+
+export const APOLLO_MOCKS: ApolloMocks = {
```
bbenoist, line 15: This block is very huge.
Maybe we could factorize things a little bit more (and avoid the need for a key-value object).

typescript/web/src/utils/tests/data.fixtures.ts
```diff
@@ -0,0 +1,69 @@
+import { Workspace, Dataset, LabelClass } from "@labelflow/graphql-types";
+
```
bbenoist, line 2: All words in the names defined in this file are inverted.

i.e. `MockDataset` should be `DatasetMock`, `MOCK_WORKSPACE_SIMPLE` should be `SIMPLE_WORKSPACE_MOCK`, etc.

typescript/web/src/utils/tests/data.fixtures.ts
```diff
@@ -0,0 +1,69 @@
...
+    {
+      id: "73a020aa-dd5e-4d6f-985a-f43d9f2744d5",
+      index: 2,
+      shortcut: "2",
+      name: "My Test Class 3 has a very long name that is way larger than anyone would ever expect for a such a field with words that do not exist with many letters such as the following one which is very loooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooong",
+      color: "#34D399",
+      labelsAggregates: {
+        totalCount: 5,
+      },
+    },
```
bbenoist, line 62: Can't we generate random values with Faker or similar?

typescript/web/src/utils/stories/apollo-decorator.tsx
```diff
@@ -0,0 +1,5 @@
+import { getMockApolloDecorator } from "../tests/mock-apollo";
+
+export const getApolloDecorator = getMockApolloDecorator;
+
+export const apolloDecorator = getMockApolloDecorator();
```
bbenoist, line 5: We don't need this file IMO

bbenoist (changes requested):

typescript/web/src/utils/tests/data.fixtures.ts
valerian: Sure I could, however I need to keep it consistent between runs, for storybook and jest snapshots.
[casual](https://www.npmjs.com/package/casual) seems to be a good alternative to Faker, and you can give it a specific random seed to keep result consistent, however if we change the mock data structure in the future it might break that.

typescript/web/src/components/dataset-classes/upsert-class-modal/upsert-class-modal.fixtures.ts
valerian: So you think I should put each mock query in its own const?

typescript/web/src/utils/tests/data.fixtures.ts
valerian: I disagree on that, see how the word is generally used in English here:
https://www.wordreference.com/enfr/mock
https://sentence.yourdictionary.com/mock

typescript/web/src/components/dataset-classes/upsert-class-modal/upsert-class-modal.fixtures.ts
bbenoist: I guess that's the only way to factorize yes. 

I see some pros of doing so:

* We'll have a more robust typing. Since in each mock declaration we would reuse the exact generated apollo types for each query or mutation.
* We can still reference the mocks from outside.
* We won't need `Object.values` to provide the array of mocks.

Do you see some cons?

typescript/web/src/utils/tests/data.fixtures.ts
bbenoist: OK I see. Thanks for the links.  
What I've asked you initially was to respect the grammar rule we use with every other variables (even the `APOLLO_MOCKS` constant is named like that). Maybe it's not the most common usage for this exact word but it's the one we use everywhere else except when we need a prefixes like the `NEXT_PUBLIC_` environment variables.
In that sense, and to respect both this rule and the links that you've given me, I think that we should use the `mockery` word if we really can't use mock as an adjective.

i.e. `MockDataset` should be `DatasetMockery`, `MOCK_WORKSPACE_SIMPLE` should be `SIMPLE_WORKSPACE_MOCKERY`, etc.

bbenoist: About the word `simple`, shouldn't it be `single` instead?

valerian: `simple` here means that there is nothing fancy, for example here `MOCK_DATASET_SIMPLE` doesn't have label classes, compared to `MOCK_DATASET_WITH_CLASSES`.

bbenoist: OK, thanks for the clarification 👍  
Simple is OK then.
If you want to disambiguate the things a little bit more when it's necessary to differentiate such variants, you can use the words `minimal` and `exhaustive`. But at this point, it's not that important...
e.g. `MINIMAL_WORKSPACE_MOCKERY` 🤷‍♂️

typescript/web/src/components/dataset-classes/upsert-class-modal/upsert-class-modal.fixtures.ts
valerian: Yeah this is better in several ways I agree, I just wanted to make sure that was what you asked

typescript/web/src/utils/tests/data.fixtures.ts
bbenoist: It's not about the amount of data we have to mock, it's about test exhaustiveness.
But it's not a priority for now. We can look at it later.

bbenoist (approved):
Code looks good to me! Very good background work @valerian 👏 
Let's proceed with merging it in #748 so that we can scale the test fixes 🚀

PR #815, opened 2022-01-20
Improve image upload performance

Jordanlelay:
## Work performed

- Update image batch size and amount for avoiding timeouts when uploading: was 2 batches of 10, now is 20 batches of 2.
- Enhance thumbnail generation performance using parallelisation and generating smaller thumbnail sizes on the largest one, instead of the original image.

## Results

We can now upload several images without experiencing timeouts

## Caveats

There is a limitation on the size of the image we can upload. I tried to upload the image `LargeImageTest.png` from our `very large images` dataset which is 23.6 Mo and it fails rather quickly with an `Array buffer allocation failed` error. 

It most likely comes from the serverless function that runs out of memory, probably when Jimp try to read the image.

The image is at a resolution of 8000x10000 (80M pixels) which should be handled by our `validateImageSize` function that is supposed to reject `.png` images bigger than 60M pixels.

### DIFF
--- a/typescript/db/src/repository/image-processing.ts
+++ b/typescript/db/src/repository/image-processing.ts
@@ -1,33 +1,40 @@
-import Blob from "fetch-blob";
-
-import { getThumbnailUrlFromImageUrl } from "@labelflow/common-resolvers/src/utils/thumbnail-url";
-import Jimp from "jimp/es";
 import type { Repository, ThumbnailSizes } from "@labelflow/common-resolvers";
+import {
+  getThumbnailUrlFromImageUrl,
+  ImageUrlInfo,
+} from "@labelflow/common-resolvers/src/utils/thumbnail-url";
+import Blob from "fetch-blob";
+import Jimp from "jimp/es";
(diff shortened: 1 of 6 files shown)


typescript/db/src/repository/image-processing.ts
```diff
@@ -103,21 +104,40 @@ export const processImage: Repository["imageProcessing"]["processImage"] =
...
-        image,
-        url,
-        putImage,
-      });
+    const { url: thumbnail500UrlNew, image: imageSrc } = thumbnail500Url
+      ? { url: thumbnail500Url, image }
+      : ((await generateThumbnail({ size: 500, image, url, putImage })) as {
+          url: string;
+          image: Jimp;
+        });
```
bbenoist: The `as` was hiding a bug...
`image` can be `undefined` as the `catch` clause of `generateThumbnail` returns `undefined`. I've fixed it during the cleanup I've pushed over your commit 😉

bbenoist (approved):

PR #826, opened 2022-01-25
Add back the last visited workspace slug cookie

### LINKED ISSUE #805: Store in a cookie the last visited Workspace ID
blocked by #843 

# Tasks

- [ ] When visiting a workspace: check if the workspace id is the same or not than the one in the cookie. If different, update it.
- [ ] When signed-in and going to labelflow.ai/
  - [ ] redirect the user to the last visited workspace or the list of workspaces if there aren't any cookies.
  - [ ] Don't redirect if it's not the same account but the same cookie
- [ ] Ignore `lastVisitedWorkspace` cookie

Jordanlelay:
## Work performed

- Add back last visited workspace slug cookie
- Add last user id cookie
- Add redirection from root page to last visited workspace
- Avoid the redirection if user has changed

## Resolved issues

closes #805


bbenoist:
Implemented by #892

PR #828, opened 2022-01-25
Add components for editing Markdown

bbenoist:


bbenoist:
While I'm still on the subject, here is a list of the integration issues I've seen so far during the design of the `MarkdownInput` component. Maybe that sharing these technical limitations could be useful to the team 🤷‍♂️

- [ ] **Monaco Editor imports some CSS files and it's not accepted by Next.js**. We have to find a solution if we want to reuse the component within LabelFlow:
  image

- [ ] I've integrated markdownlint and have it working (_error count is shown at the bottom-left of `MarkdownInput`_) but the problems aren't shown within Monaco Editor (_problems are not underlined by yellow waves, as with VS Code_).

- [ ] Code blocks have an extra new-line when used in conjunction with Markdown. As shown by it's own Storybook story, the issue does not exists when the `Code` component I've built is used in other contexts.

- [ ] Syntax hilighter styles font size are inconsistent

- [ ] Our dark theme differs a lot from the Monaco dark theme (which is neutral-gray).

- [ ] AFAIK, we don't have a way to upload other types of images than dataset-images (_to handle attachments in Markdown_)

- [ ] **Jest don't like ECMAScript modules.**  
  * I had to explicitly use the CommonJS version of the react-syntax-highlighter themes
  * I've not written the tests yet but I expect react-markdown not to work with Jest too since, IIRC, it's an ECMAScript module.

- [ ] I've started writing tests with an objective of **100% code coverage** but, as we can't manipulate the `OpenFileDialog` with `@testing-library`, it **failed** once I had to simulate a click on a `Box` whose `onClick` callback should call `inputRef.current.click()`.  
  Not that useful as we're far from 100% coverage on the project but this issue can be a tech-time subject someday: `How to mock HTMLInputComponent?`

- [ ] The **`useLintMarkdown` test** in `lint-markdown.test.tsx` is **failing**. If I understand well the problem, it looks like that `@testing-library/react-hooks` hasn't detected that a re-render actually occurred. I couldn't figure out how to fix it yet.

- [ ] **Monaco editor borders can't be rounded**. I've adapted my design so that it's barely visible. Meanwhile on GitLab, they've just decided to ignore it and let the squared Monaco overflow on the rounded borders of it's container...

- [ ] As shown last Friday, there is a **display bug in the bottom-border of Chakra tabs** when they're sticked to the tab panel. The exact same one exists in `Bootstrap` too. I'll fix it and give more details alongside the corresponding code.

Anyway, these problems can be fixed or have workarounds. I've had a lot of fun writing some of the features of `MarkdownInput` and there is some cool stuff remaining to be added 🍬  

If you want to give feedback, test the components or if you're interested in contributing, you use:
 
 ```text
 git fetch origin && git checkout feature/markdown-input && yarn && yarn storybook
 ```

PR #833, opened 2022-01-27
Update constant queries and mutations names to CONSTANT_CASE

Jordanlelay:

### DIFF
--- a/typescript/web/src/connectors/undo-store/effects/update-label.ts
+++ b/typescript/web/src/connectors/undo-store/effects/update-label.ts
@@ -2,14 +2,9 @@ import { ApolloClient, gql } from "@apollo/client";
 import { GeometryInput, Label } from "@labelflow/graphql-types";
 import { getBoundedGeometryFromImage } from "@labelflow/common-resolvers";
 import { Effect } from "..";
-import { imageDimensionsQuery } from "./shared-queries";
 
-const updateLabelMutation = gql`
-  mutation updateLabelInUndoStore(
-    $id: ID!
-    $geometry: GeometryInput
-    $labelClassId: ID
-  ) {
+const UPDATE_LABEL_MUTATION = gql`
+  mutation updateLabel($id: ID!, $geometry: GeometryInput, $labelClassId: ID) {
     updateLabel(
       where: { id: $id }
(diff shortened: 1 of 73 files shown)


bbenoist (approved):
Code looks goods to me apart from minor changes which I've already fixed when working on #834 (_a few mutations constants names with the `_QUERY` suffix)_.
Let's merge this PR into delete-service-worker right after we reintegrate #834 into this PR.

typescript/web/src/pages/[workspaceSlug]/datasets/index.tsx
```diff
@@ -24,30 +24,7 @@ import { CookieBanner } from "../../../components/cookie-banner";
...
-      labelsAggregates {
-        totalCount
-      }
-      labelClassesAggregates {
-        totalCount
-      }
-    }
-  }
-`;
+import { GET_DATASETS_QUERY } from "../../../utils/shared-queries";
```
Comment (valerian) [inline typescript/web/src/pages/[workspaceSlug]/datasets/index.tsx:28]: This query is too specific to put it in shared queries with such a generic name: it's very specifically tailored for the workspace page listing datasets. with the thumbnail of the first image of a dataset.

I think it should be renamed at least, and then I'm not sure if it should be in shared queries or not.

typescript/web/src/utils/shared-queries/index.ts
```diff
@@ -11,3 +11,27 @@ export const userQuery = gql`
     }
   }
 `;
+
+export const GET_DATASETS_QUERY = gql`
```
valerian, line 15: see previous comment

typescript/web/src/connectors/undo-store/effects/update-label.ts
```diff
@@ -31,7 +26,17 @@ const updateLabelMutation = gql`
...
+    image(where: { id: $id }) {
+      id
+      width
+      height
+    }
+  }
+`;
+
+const GET_LABEL_QUERY = gql`
   query getLabelAndGeometry($id: ID!) {
```
valerian, line 40: So you do that here, but @bbenoist undos it in #834, who is right here, and why?

valerian (changes requested):

valerian (changes requested):

PR #836, opened 2022-01-28
Fix export tests

Jordanlelay:
## Work performed

- Fix export tests

### DIFF
--- a/typescript/web/src/components/export-button/export-modal/__tests__/index.tsx
+++ b/typescript/web/src/components/export-button/export-modal/__tests__/index.tsx
@@ -1,104 +1,41 @@
 /* eslint-disable import/first */
 import { render, screen, waitFor } from "@testing-library/react";
+import { initMockedDate } from "@labelflow/dev-utils/mockdate";
 import userEvent from "@testing-library/user-event";
 import "@testing-library/jest-dom/extend-expect";
-import { ApolloProvider } from "@apollo/client";
+import { MockedProvider } from "@apollo/client/testing";
 import { ChakraProvider } from "@chakra-ui/react";
 import { PropsWithChildren } from "react";
-
-import { incrementMockedDate } from "@labelflow/dev-utils/mockdate";
-import { processImage } from "../../../../connectors/repository/image-processing";
 import { mockNextRouter } from "../../../../utils/router-mocks";
+import { BASIC_DATASET_DATA } from "../../../../utils/tests/data.fixtures";
+
+initMockedDate();
 
 mockNextRouter({
-  query: { datasetSlug: "test-dataset", workspaceSlug: "local" },
+  query: {
+    datasetSlug: BASIC_DATASET_DATA.slug,
+    workspaceSlug: BASIC_DATASET_DATA.workspace.slug,
+  },
 });
 
 import { ExportModal } from "..";
+import { getApolloMockLink } from "../../../../utils/tests/apollo-mock";
 import { theme } from "../../../../theme";
-import { client } from "../../../../connectors/apollo-client/schema-client";
-import { setupTestsWithLocalDatabase } from "../../../../utils/setup-local-db-tests";
-import {
-  CREATE_TEST_DATASET_MUTATION,
-  CREATE_TEST_IMAGE_MUTATION,
-} from "../../../../utils/tests/mutations";
-import { CREATE_LABEL_MUTATION } from "../../../../connectors/undo-store/effects/shared-queries";
-import { LabelCreateInput } from "../../../../graphql-types/globalTypes";
-
-setupTestsWithLocalDatabase();
-
-jest.mock("../../../../connectors/repository/image-processing");
-const mockedProcessImage = processImage as jest.Mock;
+import { APOLLO_MOCKS } from "../export-modal.fixtures";
 
 const wrapper = ({ children }: PropsWithChildren<{}>) => (
-  <ApolloProvider client={client}>
+  <MockedProvider link={getApolloMockLink(APOLLO_MOCKS)}>
     <ChakraProvider theme={theme} resetCSS>
       {children}
     </ChakraProvider>
-  </ApolloProvider>
+  </MockedProvider>
 );
 
-const createDataset = async (
-  datasetId = "mocked-dataset-id",
-  name = "test dataset"
-) => {
-  return await client.mutate({
-    mutation: CREATE_TEST_DATASET_MUTATION,
-    variables: { datasetId, name, workspaceSlug: "local" },
-  });
-};
-
-const labelData = {
-  geometry: {
-    type: "Polygon",
-    coordinates: [
-      [
-        [0, 0],
-        [1, 0],
-        [1, 1],
-        [0, 1],
-        [0, 0],
-      ],
-    ],
-  },
-};
-
-const createLabel = (data: LabelCreateInput) => {
-  return client.mutate({
-    mutation: CREATE_LABEL_MUTATION,
-    variables: {
-      data,
-    },
-  });
-};
-
-const imageWidth = 500;
-const imageHeight = 900;
-
-const createImage = async (name: String) => {
-  const mutationResult = await client.mutate({
-    mutation: CREATE_TEST_IMAGE_MUTATION,
-    variables: {
-      name,
-      file: new Blob(),
-      width: imageWidth,
-      height: imageHeight,
-      datasetId: "mocked-dataset-id",
-    },
-  });
-
-  const {
-    data: {
-      createImage: { id },
-    },
-  } = mutationResult;
-
-  return id;
-};
-
 test("File should be downloaded when user clicks on Export to COCO and Export", async () => {
-  await createDataset();
-  render(<ExportModal isOpen onClose={() => {}} />, { wrapper });
+  window.URL.createObjectURL = jest.fn();
+  render(<ExportModal isOpen onClose={() => {}} />, {
+    wrapper,
+  });
   const anchorMocked = {
     href: "",
     click: jest.fn(),
@@ -118,27 +55,12 @@ test("File should be downloaded when user clicks on Export to COCO and Export", 
   userEvent.click(screen.getByRole("button", { name: "Export" }));
 
   await waitFor(() => expect(anchorMocked.click).toHaveBeenCalledTimes(1));
+  await waitFor(() =>
+    expect(window.URL.createObjectURL).toHaveBeenCalledTimes(1)
+  );
 }, 20000);
 
 test("Export Modal should display the number of labels", async () => {
-  await createDataset();
-  await createDataset("second-dataset-id", "second-test-dataset");
-  mockedProcessImage.mockReturnValue({
-    width: 42,
-    height: 36,
-    mime: "image/jpeg",
-  });
-  const imageId = await createImage("an image");
-  await createLabel({
-    ...labelData,
-    imageId,
-  });
-  incrementMockedDate(1);
-  await createLabel({
-    ...labelData,
-    imageId,
-  });
-
   render(<ExportModal isOpen onClose={() => {}} />, { wrapper });
 
   await waitFor(() =>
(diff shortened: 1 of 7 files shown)


typescript/web/src/components/export-button/export-modal/__tests__/index.tsx
```diff
@@ -1,104 +1,41 @@
...
-    },
-  } = mutationResult;
-
-  return id;
-};
-
 test("File should be downloaded when user clicks on Export to COCO and Export", async () => {
-  await createDataset();
-  render(<ExportModal isOpen onClose={() => {}} />, { wrapper });
+  window.URL.createObjectURL = jest.fn();
```
bbenoist, line 35: What is this for exactly?
Also, if it's really needed, can you also make sure that the function is being called as expected?

typescript/web/src/components/export-button/export-modal/export-modal.fixtures.ts
```diff
@@ -0,0 +1,84 @@
...
+      format: ExportFormat.COCO,
+      options: {
+        coco: {
+          exportImages: false,
+          name: datasetName,
+        },
+        yolo: {
+          exportImages: false,
+          includePolygons: false,
+          name: datasetName,
```
bbenoist, line 41: `datasetName` has been set with `ExportFormat.COCO`. Isn't there a problem here?
Do you really need to use the dataset name in COCO and YOLO export options?

typescript/web/src/components/export-button/export-modal/export-modal.fixtures.ts
```diff
@@ -0,0 +1,84 @@
...
+  result: {
+    data: {
+      dataset: {
+        id: BASIC_DATASET_DATA.id,
+        imagesAggregates: {
+          totalCount: 1,
+        },
+        labelsAggregates: {
+          totalCount: 2,
+        },
```
bbenoist, line 74:
```suggestion
        imagesAggregates: { totalCount: 1 },
        labelsAggregates: { totalCount: 2 },
```

typescript/web/src/components/export-button/export-modal/utils/get-dataset-name.ts
```diff
@@ -0,0 +1,21 @@
+import { ExportFormat } from "../../../../graphql-types/globalTypes";
+
+export const getDatasetName = ({
```
bbenoist, line 3:
```suggestion
export const getDatasetExportName = ({
```

typescript/web/src/components/export-button/export-modal/utils/get-dataset-name.ts
```diff
@@ -0,0 +1,21 @@
...
+  const date = `${dateObject
+    .toLocaleDateString()
+    .split("/")
+    .reverse()
+    .join("-")}T${String(dateObject.getHours()).padStart(2, "0")}${String(
+    dateObject.getMinutes()
+  ).padStart(2, "0")}${String(dateObject.getSeconds()).padStart(2, "0")}`;
+  const datasetName = `${datasetSlug}-${format.toLowerCase()}-${date}`;
+
+  return datasetName;
```
bbenoist, line 20:
```suggestion
  return `${datasetSlug}-${format.toLowerCase()}-${date}`;
```

typescript/web/src/components/export-button/export-modal/utils/get-dataset-name.ts
```diff
@@ -0,0 +1,21 @@
...
+  format: ExportFormat;
+}) => {
+  const dateObject = new Date();
+  const date = `${dateObject
+    .toLocaleDateString()
+    .split("/")
+    .reverse()
+    .join("-")}T${String(dateObject.getHours()).padStart(2, "0")}${String(
+    dateObject.getMinutes()
+  ).padStart(2, "0")}${String(dateObject.getSeconds()).padStart(2, "0")}`;
```
bbenoist, line 17: I know that the code doesn't comes from you but these lines feel risky and complex.
If you have a few time for this, it could be nice to use some intermediary variables and eventually unit-test the `getDatasetExportName` function.

bbenoist (changes requested):

typescript/web/src/components/export-button/export-modal/export-modal.fixtures.ts
Jordanlelay: That's how it's used in the mutation, I've reproduced the same behavior

bbenoist: OK for the second question. I've written this comment before reading the parts below in the diff 😉 
But what about the first question? The YOLO name should be build with `ExportFormat.YOLO` no?

typescript/web/src/components/export-button/export-modal/__tests__/index.tsx
Jordanlelay: It's a mock for [this](https://github.com/labelflow/labelflow/blob/74253549bb57f8142177b8d8b034e417bc7d2f46/typescript/web/src/components/export-button/export-modal/export-dataset.ts#L61)

typescript/web/src/components/export-button/export-modal/export-modal.fixtures.ts
Jordanlelay: Honestly, I don't know. It does seems weird, and the opposite happens also as it is the same variable used for the both options.

I guess the Yolo options should not be passed when the export format is Coco and vice versa

bbenoist (commented):
Apart from the minor style suggestions, everything looks OK to me :)

typescript/web/src/components/export-button/export-modal/utils/get-dataset-name.ts
```diff
@@ -0,0 +1,18 @@
...
+  datasetSlug: string;
+  format: ExportFormat;
+}) => {
+  const dateObject = new Date();
+  // toLocaleDateString("fr") format is "01/11/1993"
+  const [day, month, year] = dateObject.toLocaleDateString("fr").split("/");
+  const date = `${year}-${month}-${day}`;
+  // toLocaleTimeString("fr") format is "14:31:32"
+  const time = dateObject.toLocaleTimeString("fr").replace(/:/g, "");
+  const dateAndTime = `${date}T${time}`;
```
bbenoist, line 16: Since this code uses the `fr` locale, if the end-user does not lives in France, the time (and possibly the date) would be set to an incorrect value.
As the moment lib is already needed by some of our dependencies, you can `pushd typescript/web && yarn add moment && popd` to add the package directly then use `moment().format('YYYY-MM-DDTHH:MM:SS')` instead of the whole block.

bbenoist (changes requested):

typescript/web/src/components/export-button/export-modal/utils/get-dataset-name.ts
Jordanlelay: > Since this code uses the `fr` locale, if the end-user does not lives in France, the time (and possibly the date) would be set to an incorrect value.

Are you sure about that ? My understanding of the argument `locales` is precisely to force the format to be 'fr' compliant, regardless of where you are located.

> As the moment lib is already needed by some of our dependencies, you can pushd typescript/web && yarn add moment && popd to add the package directly then use moment().format('YYYY-MM-DDThh:mm:ss') instead of the whole block.

I would strongly advocate against that lib, [as the creators themselves do](https://momentjs.com/docs/#/-project-status/), especially for such a simple usage.

bbenoist: > Are you sure about that ?

I was. But I've read the function documentation too fast and I was wrong:

> The toLocaleDateString() method returns a string with a language sensitive representation of the date portion of the specified date in the user agent's timezone.

But still, relying on this function then parsing it is not optimal compared to the previous version.

> > As the moment lib is already needed by some of our dependencies, you can pushd typescript/web && yarn add moment && popd to add the package directly then use moment().format('YYYY-MM-DDThh:mm:ss') instead of the whole block.

>
> I would strongly advocate against that lib, [as the creators themselves do](https://momentjs.com/docs/#/-project-status/), especially for such a simple usage. If we really need a lib, we can use [date-fns](https://date-fns.org/), seems better suited

OK I see... Since it's already used by the common-resolvers and release package, let's use date-fns if you don't find a vanilla alternative to the `moment.format` function 👍

Jordanlelay: done

typescript/web/src/components/export-button/export-modal/utils/get-dataset-name.ts
```diff
@@ -0,0 +1,14 @@
+import { format } from "date-fns";
+import { ExportFormat } from "../../../../graphql-types/globalTypes";
+
+export const getDatasetExportName = ({
+  datasetSlug,
+  exportFormat,
+}: {
+  datasetSlug: string;
+  exportFormat: ExportFormat;
+}) => {
```
bbenoist, line 10:
```suggestion
export const getDatasetExportName = (
  datasetSlug: string,
  exportFormat: ExportFormat,
) => {
```

typescript/web/src/components/export-button/export-modal/utils/get-dataset-name.ts
```diff
@@ -0,0 +1,14 @@
...
+  const dateAndTime = format(new Date(), "yyyy-MM-dd'T'hhmmss");
+
```
bbenoist, line 12:
```suggestion
```

typescript/web/src/components/export-button/export-modal/utils/__tests__/get-dataset-name.test.ts
```diff
@@ -0,0 +1,18 @@
...
+
+const DATASET_SLUG = "my-dataset-slug";
+const EXPORT_FORMAT = ExportFormat.COCO;
+
+describe("Get dataset export name", () => {
+  it("Returns the correct name", () => {
+    const exportName = getDatasetExportName({
+      datasetSlug: DATASET_SLUG,
+      exportFormat: EXPORT_FORMAT,
+    });
```
bbenoist, line 15:
```suggestion
    const exportName = getDatasetExportName(DATASET_SLUG, EXPORT_FORMAT);
```

typescript/web/src/components/export-button/export-modal/export-modal.fixtures.ts
```diff
@@ -0,0 +1,80 @@
...
+} from "../../../graphql-types/CountLabelsOfDatasetQuery";
+import {
+  ExportDatasetUrlQueryVariables,
+  ExportDatasetUrlQuery,
+} from "../../../graphql-types/ExportDatasetUrlQuery";
+
+const datasetName = getDatasetExportName({
+  datasetSlug: BASIC_DATASET_DATA.slug,
+  exportFormat: ExportFormat.COCO,
+});
```
bbenoist, line 22:
```suggestion
const datasetName = getDatasetExportName(
  BASIC_DATASET_DATA.slug,
  ExportFormat.COCO,
);
```

typescript/web/src/components/export-button/export-modal/export-dataset.ts
```diff
@@ -39,15 +41,10 @@ export const exportDataset = async ({
...
-    .split("/")
-    .reverse()
-    .join("-")}T${String(dateObject.getHours()).padStart(2, "0")}${String(
-    dateObject.getMinutes()
-  ).padStart(2, "0")}${String(dateObject.getSeconds()).padStart(2, "0")}`;
-  const datasetName = `${datasetSlug}-${format.toLowerCase()}-${date}`;
+  const datasetName = getDatasetExportName({
+    datasetSlug,
+    exportFormat: format,
+  });
```
bbenoist, line 47:
```suggestion
  const datasetName = getDatasetExportName(datasetSlug, format);
```

bbenoist (changes requested):

bbenoist (approved):

PR #841, opened 2022-02-01
Rework use-images-navigation hook tests

Jordanlelay:
## Work performed

- Rework use-images-navigation hook tests

### DIFF
--- a/typescript/web/src/hooks/__tests__/use-images-navigation.tsx
+++ b/typescript/web/src/hooks/__tests__/use-images-navigation.tsx
@@ -1,90 +1,49 @@
-import { ApolloProvider } from "@apollo/client";
+/* eslint-disable import/first */
 import { renderHook } from "@testing-library/react-hooks";
-import { useRouter } from "next/router";
-import { incrementMockedDate } from "../../../../dev-utils/mockdate";
-import { client } from "../../connectors/apollo-client/schema-client";
-import { processImage } from "../../connectors/repository/image-processing";
-import { setupTestsWithLocalDatabase } from "../../utils/setup-local-db-tests";
+import { mockNextRouter } from "../../utils/router-mocks";
+import { BASIC_DATASET_DATA } from "../../utils/tests/data.fixtures";
 import {
-  CREATE_TEST_DATASET_MUTATION,
-  CREATE_TEST_IMAGE_MUTATION,
-} from "../../utils/tests/mutations";
-import { useImagesNavigation } from "../use-images-navigation";
+  CURRENT_IMAGE_DATA,
+  CURRENT_NOT_IN_IMAGES_MOCKS,
+  IMAGE_1_DATA,
+  IMAGE_2_DATA,
+  IMAGE_IS_FIRST_MOCKS,
+  IMAGE_IS_LAST_MOCKS,
+  NO_IMAGES_MOCKS,
+  THREE_IMAGES_MOCKS,
+} from "../use-image-navigation.fixtures";
 
-setupTestsWithLocalDatabase();
-
-jest.mock("../../connectors/repository/image-processing");
-const mockedProcessImage = processImage as jest.Mock;
-
-const testDatasetId = "mocked-dataset-id";
-
-jest.mock("next/router", () => ({
-  useRouter: jest.fn(() => ({
-    query: { datasetSlug: "test-dataset", workspaceSlug: "local" },
-  })),
-}));
-
-async function createImage(name: String) {
-  mockedProcessImage.mockReturnValue({
-    width: 42,
-    height: 36,
-    mime: "image/jpeg",
-  });
-  const mutationResult = await client.mutate({
-    mutation: CREATE_TEST_IMAGE_MUTATION,
-    variables: {
-      file: new Blob(),
-      name,
-      datasetId: testDatasetId,
-    },
-  });
-
-  const {
-    data: {
-      createImage: { id },
-    },
-  } = mutationResult;
-
-  return id;
-}
-
-beforeEach(async () => {
-  await client.mutate({
-    mutation: CREATE_TEST_DATASET_MUTATION,
-    variables: {
-      name: "test dataset",
-      datasetId: testDatasetId,
-      workspaceSlug: "local",
-    },
-  });
+mockNextRouter({
+  isReady: true,
+  query: {
+    imageId: CURRENT_IMAGE_DATA.id,
+    datasetSlug: BASIC_DATASET_DATA.slug,
+    workspaceSlug: BASIC_DATASET_DATA.workspace.slug,
+  },
 });
 
-const Wrapper = ({ children }: React.PropsWithChildren<{}>) => (
-  <ApolloProvider client={client}>{children}</ApolloProvider>
-);
+import { useImagesNavigation } from "../use-images-navigation";
+import {
+  ApolloMockResponses,
+  getApolloMockLink,
+  getApolloMockWrapper,
+} from "../../utils/tests/apollo-mock";
+
+const renderTest = (mocks: ApolloMockResponses) => {
+  const link = getApolloMockLink(mocks);
+  const wrapper = getApolloMockWrapper(link);
+  return renderHook(() => useImagesNavigation(), { wrapper });
+};
 
 test("The currentImageIndex is undefined while loading", async () => {
-  const { result } = renderHook(() => useImagesNavigation(), {
-    wrapper: Wrapper,
-  });
+  const { result } = renderTest(NO_IMAGES_MOCKS);
 
   expect(result.current.currentImageIndex).toEqual(undefined);
 });
 
 test("The currentImageIndex is null if it can't be found in the images", async () => {
-  await createImage("image1");
-  await createImage("image2");
-  await createImage("image3");
-  (useRouter as jest.Mock).mockImplementation(() => ({
-    query: {
-      imageId: "fake-id",
-      datasetSlug: "test-dataset",
-      workspaceSlug: "local",
-    },
-  }));
-  const { result, waitForValueToChange } = renderHook(
-    () => useImagesNavigation(),
-    { wrapper: Wrapper }
+  const { result, waitForValueToChange } = renderTest(
+    CURRENT_NOT_IN_IMAGES_MOCKS
   );
 
   await waitForValueToChange(() => result.current.currentImageIndex);
@@ -93,13 +52,7 @@ test("The currentImageIndex is null if it can't be found in the images", async (
 });
 
 test("It returns the array of images when loaded", async () => {
-  await createImage("image1");
-  await createImage("image2");
-  await createImage("image3");
-  const { result, waitForValueToChange } = renderHook(
-    () => useImagesNavigation(),
-    { wrapper: Wrapper }
-  );
+  const { result, waitForValueToChange } = renderTest(THREE_IMAGES_MOCKS);
 
   await waitForValueToChange(() => result.current.images);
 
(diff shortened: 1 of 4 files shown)


typescript/web/src/hooks/__tests__/use-images-navigation.tsx
```diff
@@ -1,90 +1,58 @@
...
-const Wrapper = ({ children }: React.PropsWithChildren<{}>) => (
-  <ApolloProvider client={client}>{children}</ApolloProvider>
+const ImageNotInDatasetWrapper = getApolloMockWrapper(
+  getApolloMockLink(CURRENT_NOT_IN_IMAGES_MOCKS)
 );
 
 test("The currentImageIndex is undefined while loading", async () => {
   const { result } = renderHook(() => useImagesNavigation(), {
-    wrapper: Wrapper,
+    wrapper: NoImagesWrapper,
```
bbenoist, line 45: Instead of declaring custom wrappers for each variant, you could factorize a little bit more and have a ` renderTest` function looking like this one:

```tsx
const renderTest = (mocks: ApolloMockResponses) => {
  const link = getApolloMockLink(mocks);
  const wrapper = getApolloMockWrapper(link);
  return renderHook(() => useImagesNavigation(), { wrapper });
};
```

Then, in each test, just call something like:

```ts
const { result, waitForValueToChange } = renderTest(CURRENT_NOT_IN_IMAGES_MOCKS);
```

typescript/web/src/hooks/use-image-navigation.fixtures.ts
```diff
@@ -0,0 +1,81 @@
...
+  ApolloMockResponse,
+  ApolloMockResponses,
+} from "../utils/tests/apollo-mock";
+import { BASIC_DATASET_DATA } from "../utils/tests/data.fixtures";
+import { CURRENT_IMAGE } from "./test-utils";
+
+export const IMAGE1 = {
+  id: "929f5445-812c-4679-86bd-0f0748679f38",
+  url: "",
+  thumbnail200Url: "",
```
bbenoist, line 17: Can we at-least use something which would work?
For example something like `https://labelflow.ai/static/icon-512x512.png`.

typescript/web/src/hooks/use-image-navigation.fixtures.ts
```diff
@@ -0,0 +1,81 @@
...
+  getImagesMockGenerator(CURRENT_IMAGE_IS_LAST),
+];
+
+export const NO_IMAGES_MOCKS: ApolloMockResponses = [
+  getImagesMockGenerator([]),
+];
+
+export const CURRENT_NOT_IN_IMAGES_MOCKS: ApolloMockResponses = [
+  getImagesMockGenerator(CURRENT_IMAGE_NOT_PRESENT),
+];
```
bbenoist, line 81: Can you add an intermediary `createImagesMocks` function which would return directly the array?

typescript/web/src/hooks/use-image-navigation.fixtures.ts
```diff
@@ -0,0 +1,81 @@
...
+const CURRENT_IMAGE_IS_SECOND = [IMAGE1, CURRENT_IMAGE, IMAGE2];
+const CURRENT_IMAGE_IS_LAST = [IMAGE1, IMAGE2, CURRENT_IMAGE];
+const CURRENT_IMAGE_NOT_PRESENT = [IMAGE1, IMAGE2, IMAGE3];
+
+const getImagesMockGenerator = (
+  images: GetAllImagesOfADatasetQuery_dataset_images[]
+): ApolloMockResponse<
+  GetAllImagesOfADatasetQuery,
+  GetAllImagesOfADatasetQueryVariables
+> => {
```
bbenoist, line 42: The braces are useless here

typescript/web/src/hooks/use-image-navigation.fixtures.ts
```diff
@@ -0,0 +1,81 @@
...
+  url: "",
+  thumbnail200Url: "",
+};
+
+const CURRENT_IMAGE_IS_FIRST = [CURRENT_IMAGE, IMAGE1, IMAGE2];
+const CURRENT_IMAGE_IS_SECOND = [IMAGE1, CURRENT_IMAGE, IMAGE2];
+const CURRENT_IMAGE_IS_LAST = [IMAGE1, IMAGE2, CURRENT_IMAGE];
+const CURRENT_IMAGE_NOT_PRESENT = [IMAGE1, IMAGE2, IMAGE3];
+
+const getImagesMockGenerator = (
```
bbenoist, line 37:
```suggestion
const createImagesMock = (
```

typescript/web/src/hooks/use-image-navigation.fixtures.ts
```diff
@@ -0,0 +1,81 @@
...
+    },
+    result: {
+      data: {
+        dataset: {
+          id: BASIC_DATASET_DATA.id,
+          images,
+        },
+      },
+    },
+    nMatches: Number.POSITIVE_INFINITY,
```
bbenoist, line 59: The mocked query is not a function. Why is that needed exactly?

bbenoist (changes requested):

typescript/web/src/hooks/use-image-navigation.fixtures.ts
Jordanlelay: Are you talking about `nMatches` property ? If yes, I don't understand the `query is not a function` part

It's needed because the mutation is used in parallel. If you don't specify it, you will receive a warning saying that the mutation is not mocked after the first call

typescript/web/src/hooks/test-utils.ts
```diff
@@ -0,0 +1,12 @@
+/**
+ * This constant is not exported directly from
+ * `typescript/web/src/hooks/use-image-navigation.fixtures.ts`
+ * otherwise, the `mockNextRouter` function does not work, making all
+ * tests in `typescript/web/src/hooks/__tests__/use-images-navigation.tsx`
+ * fail
```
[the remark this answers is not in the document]
Jordanlelay: It happens in many tests, we have to mock the router before some imports. In my case, the router has to be mocked before the mocked query in the `.fixtures.ts` file.
And because I wanted to use an imported variable in my mocked router, I had to extract it from my `.fixtures.ts` file

typescript/web/src/hooks/use-image-navigation.fixtures.ts
Jordanlelay: Which array ?

bbenoist: The `[` and `]` surrounding every value declared in the block I've mentioned.

typescript/web/src/hooks/use-image-navigation.fixtures.ts
```diff
@@ -0,0 +1,77 @@
...
+export const IMAGE_3_DATA = {
+  id: "8067a965-d017-4c4b-a82d-98cc803d1020",
+  url: "https://labelflow.ai/static/icon-512x512.png",
+  thumbnail200Url: "https://labelflow.ai/static/icon-512x512.png",
+};
+
+const CURRENT_IMAGE_IS_FIRST = [CURRENT_IMAGE, IMAGE_1_DATA, IMAGE_2_DATA];
+const CURRENT_IMAGE_IS_SECOND = [IMAGE_1_DATA, CURRENT_IMAGE, IMAGE_2_DATA];
+const CURRENT_IMAGE_IS_LAST = [IMAGE_1_DATA, IMAGE_2_DATA, CURRENT_IMAGE];
+const CURRENT_IMAGE_NOT_PRESENT = [IMAGE_1_DATA, IMAGE_2_DATA, IMAGE_3_DATA];
```
bbenoist, line 35:
```suggestion

const EXAMPLE_URL = "https://labelflow.ai/static/icon-512x512.png";

const IMAGES_IDS = [
  "dd5f0305-3d69-457d-9a77-4a46cec0e10b",
  "929f5445-812c-4679-86bd-0f0748679f38",
  "51e00f79-d178-4ea1-8bb5-b0fa647f9d19",
  "8067a965-d017-4c4b-a82d-98cc803d1020",
];

export const [CURRENT_IMAGE_DATA, IMAGE_1_DATA, IMAGE_2_DATA, IMAGE_3_DATA] =
  IMAGES_IDS.map((id) => ({
    id,
    url: EXAMPLE_URL,
    thumbnail200Url: EXAMPLE_URL,
  }));

const CURRENT_IMAGE_IS_FIRST = [CURRENT_IMAGE_DATA, IMAGE_1_DATA, IMAGE_2_DATA];
const CURRENT_IMAGE_IS_SECOND = [IMAGE_1_DATA, CURRENT_IMAGE_DATA, IMAGE_2_DATA];
const CURRENT_IMAGE_IS_LAST = [IMAGE_1_DATA, IMAGE_2_DATA, CURRENT_IMAGE_DATA];
const CURRENT_IMAGE_NOT_PRESENT = [IMAGE_1_DATA, IMAGE_2_DATA, IMAGE_3_DATA];
```

typescript/web/src/hooks/__tests__/use-images-navigation.tsx
```diff
@@ -1,90 +1,49 @@
...
+import { CURRENT_IMAGE } from "../test-utils";
+
+mockNextRouter({
+  isReady: true,
+  query: {
+    imageId: CURRENT_IMAGE.id,
+    datasetSlug: BASIC_DATASET_DATA.slug,
+    workspaceSlug: BASIC_DATASET_DATA.workspace.slug,
+  },
+});
```
bbenoist, line 14:
```suggestion
```

typescript/web/src/hooks/__tests__/use-images-navigation.tsx
```diff
@@ -1,90 +1,49 @@
...
-  })),
-}));
+  CURRENT_NOT_IN_IMAGES_MOCKS,
+  IMAGE_1_DATA,
+  IMAGE_2_DATA,
+  IMAGE_IS_FIRST_MOCKS,
+  IMAGE_IS_LAST_MOCKS,
+  NO_IMAGES_MOCKS,
+  THREE_IMAGES_MOCKS,
+} from "../use-image-navigation.fixtures";
```
bbenoist, line 23:
```suggestion
  CURRENT_IMAGE_DATA,
  CURRENT_NOT_IN_IMAGES_MOCKS,
  IMAGE_1_DATA,
  IMAGE_2_DATA,
  IMAGE_IS_FIRST_MOCKS,
  IMAGE_IS_LAST_MOCKS,
  NO_IMAGES_MOCKS,
  THREE_IMAGES_MOCKS,
} from "../use-image-navigation.fixtures";

mockNextRouter({
  isReady: true,
  query: {
    imageId: CURRENT_IMAGE_DATA.id,
    datasetSlug: BASIC_DATASET_DATA.slug,
    workspaceSlug: BASIC_DATASET_DATA.workspace.slug,
  },
});
```

bbenoist (changes requested):

typescript/web/src/hooks/use-image-navigation.fixtures.ts
Jordanlelay: Looks a bit much for a test like that

bbenoist: It's common factorization... At-least it takes less space (_don't forget CURRENT_IMAGE_ 😉) and don't duplicate the URLs everywhere...

typescript/web/src/hooks/use-image-navigation.fixtures.ts
bbenoist: Yes, I'm talking about `nMatches`. Restricting a query mock to a certain number of calls adds a layer of trust. It lets you be sure that the hook don't overcall the query or that the skip property is honored...

> I don't understand the query is not a function part

Yes, I was unclear. I meant "the >mocked< query >response< is not a function". Because, the most common case where you'd want to set nMatches to `Number.POSITIVE_INFINITY` is when you've implemented the mocked query return value as a function instead of static value.

To follow-up my question, why is `GET_ALL_IMAGES_OF_A_DATASET_QUERY` be called in parallel exactly?

Jordanlelay: Actually what I said was wrong, this mock does not need this nMatches property. I was confused because of my 2 other PRs I guess 😄

bbenoist (approved):
Good job! And we end-up with less code than before!
image

PR #842, opened 2022-02-01
Fix labeling tools tests

valerian:


typescript/web/src/components/labeling-tool/image-navigation/__tests__/image-navigation-tool.tsx
```diff
@@ -1,151 +1,86 @@
...
-    push: mockedPush,
-  }));
-  const { container } = renderImageNavigationTool();
-
-  // We need to make sure that images have been loaded
-  await waitFor(() =>
-    expect(screen.getByRole("button", { name: /^Next image$/i })).toBeDefined()
+    `/${DEEP_DATASET_WITH_IMAGES_DATA.workspace.slug}/datasets/${
+      DEEP_DATASET_WITH_IMAGES_DATA.slug
+    }/images/${DEEP_DATASET_WITH_IMAGES_DATA.images[isTestNext ? 2 : 0].id}`
```
bbenoist, line 74: Can we move it into a function or at-least use intermediary variables to simplify the interpolation line?

typescript/web/src/components/labeling-tool/image-navigation/image-navigation-tool.fixtures.ts
```diff
@@ -0,0 +1,50 @@
+import { pick } from "lodash";
```
bbenoist, line 1: Isn't there a `lodash/fp` implementation?

typescript/web/src/components/labeling-tool/options-tool-bar/__tests__/edit-label-class-menu.tsx
```diff
@@ -1,167 +1,83 @@
...
-      })
-    );
-  });
-
-  // The id is generated by the effect. We retrieve it from the first mutation
-  // which is "createLabelClass"
-  const { id } = (client.mutate as jest.Mock).mock.calls[0][0].variables.data;
+  const mockLink = getApolloMockLink(APOLLO_MOCKS);
+  renderEditLabelClassMenu(mockLink);
+  act(() =>
```
bbenoist, line 57: Shouldn't you `await` the `act`?

typescript/web/src/components/labeling-tool/options-tool-bar/__tests__/edit-label-class-menu.tsx
```diff
@@ -1,167 +1,83 @@
...
-    mutation: CREATE_LABEL_CLASS_MUTATION,
-    variables: {
-      id: "existing label class id",
-      name: "existing label class",
-      color: "0xaa45f7",
-      datasetId: testDatasetId,
-    },
-  });
+beforeEach(() => {
+  act(() =>
```
bbenoist, line 44: Shouldn't you `await` the `act`?

typescript/web/src/components/labeling-tool/options-tool-bar/__tests__/edit-label-class-menu.tsx
```diff
@@ -1,167 +1,83 @@
...
-          },
-          where: { id: "my label id" },
-        },
-      })
-    );
-  });
+  userEvent.click(
+    screen.getByText(DEEP_DATASET_WITH_CLASSES_DATA.labelClasses[1].name)
+  );
+  // Intentionally done twice
```
bbenoist, line 79: Why?

typescript/web/src/components/labeling-tool/options-tool-bar/options-tool-bar.fixtures.ts
```diff
@@ -0,0 +1,208 @@
...
+              "name",
+              "color"
+            )
+          : {
+              id: variables.id,
+              name: "uknown label class name",
+              color: "#123456",
+            },
+    },
+  })),
```
bbenoist, line 119: This is starting to be complex and non-factorized (the `[1]` should be done only once). Can you move it into its own function(s)?

typescript/web/src/components/labeling-tool/options-tool-bar/options-tool-bar.fixtures.ts
```diff
@@ -0,0 +1,208 @@
...
+  },
+  result: jest.fn(({ data: { id, name, color } }) => ({
+    data: {
+      createLabelClass: {
+        id: id ?? uuid(),
+        name,
+        color,
+      },
+    },
+  })),
```
bbenoist, line 157: Again, I think that it should be moved into its own function(s) but this one is less critical.

typescript/web/src/components/labeling-tool/options-tool-bar/options-tool-bar.fixtures.ts
```diff
@@ -0,0 +1,208 @@
...
+  },
+  nMatches: Number.POSITIVE_INFINITY,
+  result: jest.fn((variables) => ({
+    data: {
+      updateLabel: {
+        id: variables.where.id,
+        labelClass: { id: variables.data.labelClassId! },
+      },
+    },
+  })),
```
bbenoist, line 197: Even less critical for this one indeed.

typescript/web/src/utils/tests/data.fixtures.ts
```diff
@@ -1,17 +1,29 @@
...
   workspace: WorkspaceData;
 };
 
+export type ImageData = SetRequired<
+  Pick<Image, "id" | "name" | "url" | "thumbnail200Url">,
+  "thumbnail200Url"
+> & {
+  dataset: DatasetData;
+};
+
```
bbenoist, line 26: And here is where trying to base our fixtures on a `Pick` of global data types instead of either the complete types or the generated apollo client types is failing IMO. You have to work-around cross-references and omit or redeclare properties.

typescript/web/src/utils/tests/data.fixtures.ts
```diff
@@ -31,6 +47,46 @@ export const BASIC_DATASET_DATA: DatasetData = {
   slug: "my-test-dataset",
   workspace: BASIC_WORKSPACE_DATA,
   labelClasses: [],
+  images: [
+    {
```
bbenoist, line 51: In #841 @Jordanlelay created some data which you could reuse (and eventually make it use your URLS which are closer to the reality).

typescript/web/src/components/labeling-tool/image-navigation/image-navigation-tool.fixtures.ts
```diff
@@ -0,0 +1,50 @@
...
+
+export const GET_ALL_IMAGES_OF_A_DATASET_MOCK: ApolloMockResponse<
+  GetAllImagesOfADatasetQuery,
+  GetAllImagesOfADatasetQueryVariables
+> = {
+  request: {
+    query: GET_ALL_IMAGES_OF_A_DATASET_QUERY,
+    variables: MATCH_ANY_PARAMETERS,
+  },
+  result: jest.fn((variables) => ({
```
bbenoist, line 30: Again here, this should be moved in function(s) so that the mock feels less complex to read

typescript/web/src/components/labeling-tool/options-tool-bar/__tests__/edit-label-class-menu.tsx
valerian: only when the act wraps a promise returning function :)

typescript/web/src/utils/tests/data.fixtures.ts
valerian: why not the other way around? wouldn't it make more sense that `use-image-navigation.fixtures.ts` imports all from `data.fixtures.ts`, since it's already importing `BASIC_DATASET_DATA`?

bbenoist: Indeed it has to be placed at the lowest level, yes 😉

typescript/web/src/components/labeling-tool/image-navigation/__tests__/image-navigation-tool.tsx
```diff
@@ -1,151 +1,89 @@
...
   renderImageNavigationTool();
-
   await waitFor(() =>
     expect(screen.queryByDisplayValue(/1/i)).toBeInTheDocument()
   );
   await waitFor(() => expect(screen.queryByText(/1/i)).toBeInTheDocument());
 });
 
-test("should select previous image when the left arrow is pressed", async () => {
+const testNextPrevImage = (isTestNext: boolean) => async () => {
```
bbenoist, line 52: To avoid issues, I'd have used `position: "previous" | "next"` instead.

typescript/web/src/components/labeling-tool/image-navigation/image-navigation-tool.fixtures.ts
```diff
@@ -0,0 +1,54 @@
...
+  DEEP_DATASET_WITH_IMAGES_DATA,
+  BASIC_DATASET_DATA,
+} from "../../../utils/tests/data.fixtures";
+
+const DATASET_SLUG_DATA = {
+  [BASIC_DATASET_DATA.slug]: BASIC_DATASET_DATA,
+  [DEEP_DATASET_WITH_IMAGES_DATA.slug]: DEEP_DATASET_WITH_IMAGES_DATA,
+};
+
+const GetAllImagesOfADatasetMockResult = jest.fn(
```
bbenoist, line 22:
```suggestion
const getAllImagesOfADatasetMockResult = jest.fn(
```

typescript/web/src/components/labeling-tool/options-tool-bar/__tests__/edit-label-class-menu.tsx
```diff
@@ -1,167 +1,81 @@
...
+  const mockLink = getApolloMockLink(APOLLO_MOCKS);
+  renderEditLabelClassMenu(mockLink);
+  act(() =>
+    userEvent.type(screen.getByPlaceholderText(/Search/), "newClassTest")
+  );
+  await waitFor(() => expect(screen.getByText("Create class")).toBeDefined());
+  userEvent.click(screen.getByText("Create class"));
+  await act(() => mockLink.waitForAllResponsesRecursively());
+  expect(CREATE_LABEL_CLASS_ACTION_MOCK.result).toHaveBeenCalled();
+  expect(UPDATE_LABEL_CLASS_ACTION_MOCK.result).toHaveBeenCalled();
```
bbenoist, line 63: Why is the update mock called when creating a new entity?

typescript/web/src/components/labeling-tool/options-tool-bar/__tests__/edit-label-class-menu.tsx
```diff
@@ -1,167 +1,81 @@
...
 
 it("should change a class", async () => {
-  renderEditLabelClassMenu();
-
+  const mockLink = getApolloMockLink(APOLLO_MOCKS);
+  renderEditLabelClassMenu(mockLink);
   await waitFor(() =>
-    expect(screen.getByText(/existing label class/)).toBeDefined()
+    expect(
+      screen.getByText(DEEP_DATASET_WITH_CLASSES_DATA.labelClasses[1].name)
```
bbenoist, line 71: You should move `DEEP_DATASET_WITH_CLASSES_DATA.labelClasses[1]` in a variable or directly use the correct LabelClass fixture if it exists (otherwise extract it).

typescript/web/src/utils/tests/data.fixtures.ts
```diff
@@ -21,6 +26,55 @@ export type LabelClassData = Pick<
...
+  Pick<Image, "id" | "name" | "url" | "thumbnail200Url" | "width" | "height">,
+  "thumbnail200Url"
+> & {
+  dataset: DatasetData;
+  labels: (Omit<LabelData, "imageId" | "labelClass"> & {
+    labelClass: Omit<LabelClassData, "dataset">;
+  })[];
+};
+
+const BASIC_LABEL_GEOMETRY_PARTIAL = {
```
bbenoist, line 57: Why is there a `BASIC_` prefix?  
What makes the data being Partial?  
Why is there no TypeScript type?

typescript/web/src/utils/tests/data.fixtures.ts
bbenoist: You've marked the comment as resolved by yourself but I couldn't validate the changes before and I see no constant like `IMAGE_1_DATA` being reused here.

typescript/web/src/utils/tests/data.fixtures.ts
```diff
@@ -71,9 +177,83 @@ export const DEEP_DATASET_WITH_CLASSES_DATA: DatasetData = {
       },
     },
   ],
+  images: [],
+};
+
+const NESTED_LABEL_CLASS_IN_DATASET_WITH_LABELS_DATA: Omit<
```
bbenoist, line 183: Maybe that it's something that should rather been done in the next "test-cleanup PR" but I feel like with these things we're going too far away from the idea to have something reusable everywhere.
Can't we move such cases (relations data) in the query mocks instead?

bbenoist (changes requested):

typescript/web/src/components/labeling-tool/options-tool-bar/__tests__/edit-label-class-menu.tsx
valerian: the problem is that the queries themselves are not really well named, the first query creates a LabelClass, the second query updates the LabelClass (or just "class") of a Label.

typescript/web/src/utils/tests/data.fixtures.ts
valerian: I closed it because the situation shifted, since you seemed to agree to my proposition of basing on `data.fixtures.ts` instead of `use-image-navigation.fixtures.ts` as the mock data source for images.

So I assumed you would make a comment on the changes I made to `use-image-navigation.fixtures.ts` instead.

typescript/web/src/utils/tests/data.fixtures.ts
valerian: Yeah, now it would probably make more sense to stop having relations in `data.fixtures.ts` (and maybe provide some helper function to create relations, I will see if it seems useful or not).

typescript/web/src/components/labeling-tool/options-tool-bar/__tests__/edit-label-class-menu.tsx
valerian: This should be fixed with the refactor of `data.fixtures.ts` that you suggested in your other comment, in order to remove relations/nested data.

typescript/web/src/components/labeling-tool/options-tool-bar/__tests__/edit-label-class-menu.tsx
bbenoist: OK I see... Let's keep it as-is for now. We'll eventually cleanup this part in the longer term...

PR #843, opened 2022-02-01
Merge user and workspace list

bbenoist:
## Work performed

* Cleanup test code
* Add a simplified test wrapper
* Fix WorkspaceNameInput and CreateWorkspaceModal component tests
* Fix most of the Storybook and Cypress tests
* Merge multiple branches waiting for tests to be fixed


bbenoist:
@valerian @Jordanlelay as agreed, you can review the 2 last commits I've pushed which corresponds to the 2 tasks in the above list. The 3rd one didn't need changes as optional is still useful at some places (and will be needed with public datasets anyway)

PR #845, opened 2022-02-02
Use https://docs.labelflow.ai as documentation URL

bbenoist:
## Work performed

I've made the code use `https://docs.labelflow.ai` instead of the gitbook URL.

## Results

The links in our app should now use `https://docs.labelflow.ai`.

## Problems encountered

The URL wasn't declared in a constant.

## Caveats

The documentation URL don't points to a specific version of the app. But it's not really a problem for now since we're not supporting on-premise.

## Resolved issues

No

## Newly raised issues

No


typescript/web/src/constants.ts
```diff
@@ -2,3 +2,6 @@ export const WEB_APP_URL_ORIGIN =
   process.env.NEXTAUTH_URL ??
   (process.env.VERCEL_URL && `https://${process.env.VERCEL_URL}`) ??
   "https://labelflow.ai";
+
+/** Public documentation URL */
+export const DOCUMENTATION_URL = "https://docs.labelflow.ai/";
```
Jordanlelay, line 7: Going to this URL provokes a redirection to `https://docs.labelflow.ai/labelflow` so shouldn't we include the `/labelflow` part in it ? It can be misleading

bbenoist: Well, I see at-least 2 reasons for not using the `/labelflow` suffix:
* Responsibility repartition. The `/labelflow` route is a gitbook internal. Whereas `docs.labelflow.ai` will always redirect to the LabelFlow documentation, regardless of its hosting provider.
* User simplification: whenever possible, we try to refer to the shortest URLs versions. As for Chrome which grays-out the path in URLs for example.

Jordanlelay (approved):

typescript/web/src/constants.ts
Jordanlelay: disagree with the second point as I was referring to the redirection, but let's merge it as is

Issue #646, opened 2021-12-09
Fix post issue with URLs


etiennedupont:
Won't it be safer to replace `article?.image?.url !== null` in line 36 by just `article?.image?.url` instead of ignoring `undefined` by using `?` in line 39 (` url: article?.image?.url,`)?

_Originally posted by @bbenoist in https://github.com/labelflow/labelflow/pull/642#r766079807_

PR #849, opened 2022-02-03
Only build the working Storybook files until the service worker is fully dropped

bbenoist:
## Work performed

I've listed the storybook directories and only included the ones which were not causing the Storybook build to fail.

## Results

Storybook should work again.

## Problems encountered

Storybook does not print which story is causing the build to fail.

## Caveats

* Each `__stories__` directory is explicitly listed in the `.storybook/main.js` file.
* We'll have to revert the changes once the stories gets fixed.

## Resolved issues

No but it contributes to #848 

## Newly raised issues

> **TODO**: Add task "Revert d320ec6855c2cfc5fd784f859e96ff4b02ed62a1" at the end of #646

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,6 +58,8 @@ jobs:
         with:
           path: "**/node_modules"
           key: ${{ runner.os }}-modules-${{secrets.ACTIONS_CACHE_VERSION}}-${{ hashFiles('yarn.lock') }}
+      - name: Generate code
+        run: yarn codegen
       - name: Lint project code
         run: yarn lint:ts
 
@@ -156,6 +158,8 @@ jobs:
         run: cd typescript/db && yarn deploy && yarn generate
         env:
           POSTGRES_EXTERNAL_URL: postgresql://admin:admin@localhost:5433/labelflow?schema=public
+      - name: Generate code
+        run: yarn codegen
       - name: run tests
         run: yarn test
         env:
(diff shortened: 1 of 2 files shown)


.storybook/main.js
```diff
@@ -1,29 +1,66 @@
-const NodePolyfillPlugin = require("node-polyfill-webpack-plugin")
+const NodePolyfillPlugin = require("node-polyfill-webpack-plugin");
+
+// FIXME SW Delete this constant once every story has been fixed following the service worker removal
+const WORKING_STORIES = [
```
Jordanlelay, line 4: Since you've declared this constant as "working", we could extract the non working in their own variable
Or just name the constant `STORIES`

.github/workflows/ci.yml
```diff
@@ -58,6 +58,8 @@ jobs:
         with:
           path: "**/node_modules"
           key: ${{ runner.os }}-modules-${{secrets.ACTIONS_CACHE_VERSION}}-${{ hashFiles('yarn.lock') }}
+      - name: Generate code
```
Jordanlelay, line 61: Could we use something more precise? Like 
`Generate graphql schema, mutations and query types`

.storybook/main.js
bbenoist: It's just a temporary list where broken stories are commented so you just have to uncomment them and start fixing. 
As the `FIXME SW` comment above implies, it will never make its way into main. But I've rename it into `STORIES` anyway so that we can go forward. Thanks for the suggestion!

.github/workflows/ci.yml
bbenoist: As long as the script being called is named `codegen`, I'd prefer the name to match it.
Since you've asked for some sort of disambiguation, I have modified an existing line calling `codegen:ci` however to denote the fact that it also checks for uncommitted changes after generating the source code.

bbenoist:
CI seems OK for Storybook & Chromatic!

image


It looks like they are some display issues in some of the working stories but we'll fix them on #748 rather than this PR...

Jordanlelay (approved):

PR #863, opened 2022-02-08
Fix timeout on label export

### LINKED ISSUE #844: Export labels: fix timeout on COCO and YOLO json export
## Tasks
- Log why there is a timeout on a medium size dataset
- Try to use Prisma relation to batch every DB call in a single query

### Yolo:
- For every image, it queries the list of LabelClasses
- Action batch the LabelClasses of a dataset

### Coco:
- For every label, it queries the image dimensions
- Action: query the image dimension only once

Jordanlelay:
## Work performed

- Rework the way labels are exported
- Remove unused flickr_url

## Results

The labels, labelClasses and images of a dataset are now queried once for all at the beginning of the export procedure. A temporary object is built for fast access when iterating over the labels / images.

Tests I've conducted over a 11000+ labels dataset:
- yolo export -> From ~20s to ~10s
- coco export -> From `timeout` to ~8s

## Caveats

The tests were **only** focusing on labels and not images. If too much timeouts because of images happen, there is another way we could potentially improve performances:

- In both COCO and YOLO exports, for each image, this query is executed: `await repository.upload.get(image.url, req);`, returning an `ArrayBuffer` of the to-be-exported image. We could potentially batch these calls as we already have the full list of the dataset's images.

## Resolved issues

Closes #844

### DIFF
--- a/typescript/common-resolvers/src/export/format-coco/add-image-dimensions-to-labels.ts
+++ b/typescript/common-resolvers/src/export/format-coco/add-image-dimensions-to-labels.ts
@@ -1,25 +1,15 @@
 import { DbLabel, DbImage } from "../../types";
 
-type ImageDimensions = {
-  width: number;
-  height: number;
-};
+type ImageDimensions = Pick<DbImage, "width" | "height">;
 
 export const addImageDimensionsToLabels = (
   labels: DbLabel[],
   images: DbImage[]
 ) => {
-  const dimensionsByImage = images.reduce(
-    (acc: { [imageId: string]: ImageDimensions }, image) => {
-      const key = image.id;
-      if (!acc[key]) {
-        acc[key] = { width: image.width, height: image.height };
-      }
-      return acc;
-    },
+  const dimensionsByImage = images.reduce<Record<string, ImageDimensions>>(
+    (prev, { id, width, height }) => ({ ...prev, [id]: { width, height } }),
     {}
   );
-
   return labels.map((label) => ({
     ...label,
     imageDimensions: dimensionsByImage[label.imageId],
--- a/typescript/common-resolvers/src/export/format-coco/add-image-dimensions-to-labels.ts
+++ b/typescript/common-resolvers/src/export/format-coco/add-image-dimensions-to-labels.ts
@@ -1,5 +1,4 @@
-import { DbImage } from "../..";
-import { DbLabel } from "../../types";
+import { DbLabel, DbImage } from "../../types";
 
 type ImageDimensions = {
   width: number;
--- a/typescript/common-resolvers/src/export/format-coco/add-image-dimensions-to-labels.ts
+++ b/typescript/common-resolvers/src/export/format-coco/add-image-dimensions-to-labels.ts
@@ -1,22 +1,28 @@
-import { DbLabel, Repository } from "../../types";
-import { DbLabelWithImageDimensions } from "./coco-core/types";
+import { DbImage } from "../..";
+import { DbLabel } from "../../types";
 
-export const addImageDimensionsToLabels = async (
+type ImageDimensions = {
+  width: number;
+  height: number;
+};
+
+export const addImageDimensionsToLabels = (
   labels: DbLabel[],
-  repository: Repository,
-  user?: { id: string }
-): Promise<DbLabelWithImageDimensions[]> => {
-  return await Promise.all(
-    labels.map(async (label) => {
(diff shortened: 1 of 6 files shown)


typescript/common-resolvers/src/export/format-yolo/export-to-yolo.ts
```diff
@@ -6,6 +6,19 @@ import { getImageName } from "../common";
...
+        const key = label.imageId;
+        if (!acc[key]) {
+          acc[key] = [];
+        }
+        acc[key].push(label);
+        return acc;
+      }, {})
+    : {};
+};
+
```
bbenoist, line 21: You can use [lodash groupBy](https://lodash.com/docs/#groupBy) function to do this operation.
I guess you'd end-up with something like:

```suggestion
import { isEmpty, groupBy } from "lodash/fp";
import { ExportFunction } from "../types";

const groupLabelsByImage = (labels: DbLabel[]) =>
  isEmpty(labels) ? {} : groupBy(labels, "imageId");
```

typescript/common-resolvers/src/export/format-coco/add-image-dimensions-to-labels.ts
```diff
@@ -1,22 +1,28 @@
-import { DbLabel, Repository } from "../../types";
-import { DbLabelWithImageDimensions } from "./coco-core/types";
+import { DbImage } from "../..";
+import { DbLabel } from "../../types";
 
-export const addImageDimensionsToLabels = async (
+type ImageDimensions = {
```
bbenoist, line 4: If you're OK, let's rename and move the type into a `geometry` directory of @labelflow/utils. We will populate the directory with all geometry-related types and functions and eventually switch easier to a dedicated library if it becomes too big..
No need to make sure it's used everywhere for now but it will let us reuse it anywhere a 2D size is needed 🚀 

```suggestion
export type Size2D = {
```

typescript/common-resolvers/src/export/format-coco/add-image-dimensions-to-labels.ts
```diff
@@ -1,22 +1,28 @@
...
-      if (image == null) {
-        throw new Error(`Missing image with id ${imageId}`);
+  images: DbImage[]
+) => {
+  const dimensionsByImage = images.reduce(
+    (acc: { [imageId: string]: ImageDimensions }, image) => {
+      const key = image.id;
+      if (!acc[key]) {
+        acc[key] = { width: image.width, height: image.height };
       }
```
bbenoist, line 18:
```suggestion
  const dimensionsByImage = images.reduce<Record<string, Size2D>>(
    (prev, { id, width, height }) => ({ ...prev, [id]: { width, height }}),
    {},
  );
```

typescript/common-resolvers/src/export/format-coco/add-image-dimensions-to-labels.ts
```diff
@@ -1,22 +1,28 @@
...
-      return {
-        ...label,
-        imageDimensions: { height: image.height, width: image.width },
-      };
-    })
+      return acc;
+    },
+    {}
   );
```
bbenoist, line 22:
```suggestion
```

typescript/common-resolvers/src/export/format-coco/add-image-dimensions-to-labels.ts
```diff
@@ -1,22 +1,28 @@
...
+
```
bbenoist, line 23:
```suggestion
```

bbenoist (changes requested):

bbenoist:
I've made a review but it'd like @valerian to give it a look too.

nicOwlas:
I can see the improvement on the Georgia test dataset. It's much faster to export 👍

typescript/common-resolvers/src/export/format-coco/add-image-dimensions-to-labels.ts
Jordanlelay: Actually it's a small part of the `DbImage` type so maybe I could just `Pick` from it to reflect that it doesn't come out of nowhere

bbenoist: Yes, let's pick the type for now, it's simpler 👍

bbenoist (approved):

PR #864, opened 2022-02-08
Change announcements button icon from BsMegaphone to IoMegaphoneOutline

bbenoist:
## Work performed

Current icon line width is too small so I've changed the icon

## Results

Icon is expected to fit better with the other icons

## Problems encountered

No

## Caveats

That's subjective but I preferred the design previous version 😿 

## Resolved issues

@nicOwlas feel free to create one if you want to track your request

## Newly raised issues

No

### DIFF
--- a/typescript/web/src/components/announcements/announcements-button.tsx
+++ b/typescript/web/src/components/announcements/announcements-button.tsx
@@ -1,6 +1,6 @@
 import { Box, Circle } from "@chakra-ui/react";
 import { isNil } from "lodash/fp";
-import { BsMegaphone } from "react-icons/bs";
+import { IoMegaphoneOutline } from "react-icons/io5";
 import { IconButton } from "../icon-button";
 import { Announcements, AnnouncementsProps } from "./announcements";
 import { useAnnouncements } from "./announcements.context";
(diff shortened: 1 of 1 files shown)


valerian (approved):
Code is approved, but I don't really like the new icon

nicOwlas:
Lets not release this PR. The goal was to see if another react-icon had a better line stroke matching our other icons to have a quick win. It's not the case (and I agree that present icon is nicer).

We'll do another issue with a proper icon (the one we have right now but with a larger line stroke) and the megaphone will be moved inside the help menu
image

bbenoist:
@valerian Thanks for the review and the honest opinion about the icon!
This PR was a quick edit directly on GitHub just to see if the icon thickness was matching the other ones we use 

We'll discuss the issue in next DSM but we could:
* [Configure webpack to include SVG files](https://blog.logrocket.com/how-to-use-svgs-in-react/)
* Use a customized version made by @nicOwlas of the initial icon
* Give credits to the original icon we've customized where appropriate

bbenoist:
@nicOwlas I think that we need to discuss about moving the icon into the user menu as I disagree with it.

IMO it will hide the notification for new features as they are expected to be without booster bars when being logged-in. They will end-up being missed as the icon will be hidden in a menu that no one actually use frequently.

nicOwlas:
We would put a red dot on the `?` menu icon and megaphone to notify on new announcements. It's the notification bell which should be in the header bar (we can keep the megaphone there so long we don't have any notification bell)

bbenoist:
> we can keep the megaphone there so long we don't have any notification bell

👍 This is what we initially agreed (and why it was a notification bell in the mockups). I've just used a megaphone because it made more sense as-long-as the notifications are only announcements.

PR #867, opened 2022-02-08
WIP undo store tests: create label class and update label

valerian:


typescript/web/src/connectors/undo-store/effects/create-label-class-and-update-label.test.tsx
```diff
@@ -1,159 +1,101 @@
...
-    );
+    expect(CREATE_LABEL_CLASS_ACTION_MOCK.result).toHaveBeenCalledWith({
+      data: expect.objectContaining({
+        name: NEW_LABEL_CLASS_DATA.name,
+        color: NEW_LABEL_CLASS_DATA.color,
+        datasetId: DEEP_DATASET_WITH_CLASSES_DATA.id,
+      }),
+    });
+    const newLabelClassId = (CREATE_LABEL_CLASS_ACTION_MOCK.result as jest.Mock)
+      .mock.calls[0][0].data.id;
```
bbenoist, line 54: As you've previously done in another PR, the result `jest.fn` declaration should be extracted in a dedicated const in the fixture, so that you can directly reuse it and avoid the need to cast the `result` as `jest.Mock`.

i.e.:

```suggestion
    const [{data: {id: newLabelClassId]}] = createLabelClassActionMockResult.mock.calls[0];
```
(_Maybe it's not extractable in a one-liner because of nullable types but you get the idea_)

typescript/web/src/connectors/undo-store/effects/label-and-label-class.fixtures.ts
```diff
@@ -0,0 +1,140 @@
...
+} from "../../../utils/fixtures";
+import {
+  CREATE_LABEL_CLASS_QUERY,
+  DELETE_LABEL_CLASS_MUTATION,
+  GET_LABEL_QUERY,
+  UPDATE_LABEL_MUTATION as UPDATE_LABEL_MUTATION_SHARED,
+} from "./shared-queries";
+import { UPDATE_LABEL_MUTATION as UPDATE_LABEL_CLASS_OF_LABEL_MUTATION } from "./update-label-class-of-label";
+
+const [PREVIOUS_LABEL_CLASS_DATA] = DEEP_DATASET_WITH_CLASSES_DATA.labelClasses;
```
bbenoist, line 36: Can't we avoid this by extracting the inner label-classes of  `DEEP_DATASET_WITH_CLASSES_DATA` in `*_LABEL_CLASS_DATA` constants?

(_please update your branch with the latest revision of #843 if you decide to do it 🙏_)

typescript/web/src/connectors/undo-store/effects/label-and-label-class.fixtures.ts
```diff
@@ -0,0 +1,140 @@
...
+
+const createLabelClassActionMockResult = jest.fn(
+  (
+    variables: CreateLabelClassActionMutationVariables
+  ): FetchResult<CreateLabelClassActionMutation> => ({
+    data: {
+      createLabelClass: {
+        id: variables.data.id ?? uuid(),
+        name: variables.data.name,
+        color: variables.data.color ?? "#123456",
```
bbenoist, line 64: Fun-fact: `#123456` is a nice color :trollface: 
image

bbenoist (changes requested):

typescript/web/src/connectors/undo-store/effects/label-and-label-class.fixtures.ts
valerian: I agree with that and as I said in a comment from my previous PR this should be done as a separate task, since it involves quite a lot of changes to all the tests and fixtures files.

I _could_ do it in this PR though, since it's one of the last in the drop jest tests collection.

bbenoist:
Cherry-picked in #843

PR #871, opened 2022-02-10
Add temporary modal for migrating local datasets

### LINKED ISSUE #868: Display on local/datasets a modal alerting the datasets will be lost if not transfered next Monday
## Task

- [ ] Display an alert on the local workspace / dataset to warn users their dataset would be deleted

Jordanlelay:
## Work performed

- Add a modal on dataset's local workspace page for migrating to online workspaces

## Resolved issues

Closes #868


typescript/web/src/pages/[workspaceSlug]/datasets/index.tsx
```diff
@@ -41,6 +54,62 @@ export const getDatasetsQuery = gql`
...
 `;
 
+const MigrateLocalDatasetsModal = () => {
+  const router = useRouter();
+  const workspaceSlug = router?.query.workspaceSlug as string;
+  const [isLocalDatasetsModalOpen, setIsLocalDatasetsModalOpen] =
+    useState<boolean>(false);
+  const closeModal = useCallback(() => {
+    setIsLocalDatasetsModalOpen(false);
+  }, []);
```
Comment (bbenoist) [inline typescript/web/src/pages/[workspaceSlug]/datasets/index.tsx:64]: For such cases, you can use the `useBoolean` hook from Chakra 😉

typescript/web/src/pages/[workspaceSlug]/datasets/index.tsx
```diff
@@ -41,6 +54,62 @@ export const getDatasetsQuery = gql`
...
+      isOpen={isLocalDatasetsModalOpen}
+      onClose={closeModal}
+      isCentered
+      size="xl"
+    >
+      <ModalOverlay />
+      <ModalContent>
+        <ModalCloseButton />
+        <ModalHeader>
+          <Heading as="h2" size="md">
```
Comment (bbenoist) [inline typescript/web/src/pages/[workspaceSlug]/datasets/index.tsx:81]: I wonder why you set both `h2` and the `size` 🤔

bbenoist (commented):
My remarks are not blocker, code is not perfect but it's OK if we consider that the commit will be reverted in 1 week.
Let's see the checks results before approving the PR however.

typescript/web/src/pages/[workspaceSlug]/datasets/index.tsx
Comment (Jordanlelay) [reply typescript/web/src/pages/[workspaceSlug]/datasets/index.tsx:81]: `as` is more a semantically prop to make appear a `<h2>...</h2>` in the DOM whereas the `size` is really to control how big/small it will visually appear to the user.

**However**, `Heading` render `h2` by default so I will remove it 😛

bbenoist (changes requested):
Cypress is broken as you expected

bbenoist (changes requested):
When clicking on "Try it now", the warning shows up: 
image

Jordanlelay:
> When clicking on "Try it now", the warning shows up

Because you are on the `local` workspace slug right ?

bbenoist:
yes

Jordanlelay:
Isn't it as expected then ?

bbenoist:
I'd have shown the warning _after_ the welcome modal has been closed and _only if_ `hasUserTriedApp` was `false` when loading the page.

bbenoist (changes requested):
The modal is shown even if the only dataset it the tutorial one.

Also there is the case below, but we'll keep it as-is for now since it's the correct modal that's shown first and the warning will still be shown after the update and reload:

image

bbenoist (approved):

bbenoist (approved):

PR #874, opened 2022-02-10
Fix dark mode color problem in invitation modal and user menu

### LINKED ISSUE #861: Invitation acceptation component background is white under dark mode
## Describe the bug

Invitation acceptation component background is white under dark mode

## To Reproduce

1. Enable dark mode
2. Accept an invite

## Expected behavior

Message should be readable

## Screenshots

image

## Device information

 

- **Device:** PC
- **OS:** Windows 11
- **Browser:** MS Edge
- **Version:** 98

## Additional context

No

Jordanlelay:
## Work performed

- Handle dark mode in invitation modal
- Handle dark mode in invalid invitation modal
- Extract component from duplicated code
- Fix avatar color icon in dark mode

## Resolved issues

Closes #861


typescript/web/src/components/invitation-manager/invitation-container.tsx
```diff
@@ -0,0 +1,56 @@
...
+type CompleteProps = PropsWithChildren<InvitationContainerProps>;
+
+export const InvitationContainer = ({
+  header,
+  details,
+  children,
+}: CompleteProps) => {
+  return (
+    <Center h="full">
+      <chakra.div
```
bbenoist, line 18: While you're at it, can we switch to standard Chakra components?

Jordanlelay: I can but what's wrong with that ?

bbenoist: It add extra boilerplate code and complexifies what should have just been an imbrication of VStack, Header and HStack without fixed fontSize, fontWeight, backgrounds, some of the paddings, etc.

bbenoist: It's not critical but while there is a few dev bandwidth while waiting for SW to be finished, I think that we can take a few time to clean this up 🧹

typescript/web/src/components/invitation-manager/invitation-container.tsx
```diff
@@ -0,0 +1,44 @@
...
+  details: string;
+};
+
+type CompleteProps = PropsWithChildren<InvitationContainerProps>;
+
+export const InvitationContainer = ({
+  header,
+  details,
+  children,
+}: CompleteProps) => {
```
bbenoist, line 21: Braces are useless now

typescript/web/src/components/invitation-manager/invitation-container.tsx
```diff
@@ -0,0 +1,44 @@
...
+        borderRadius="md"
+        boxShadow="lg"
+        maxW="2xl"
+        my="3.75rem"
+      >
+        <Heading fontSize="xl" fontWeight="semibold">
+          {header}
+        </Heading>
+        <Flex>{details}</Flex>
+        <Flex w="100%" justify="flex-end">
```
bbenoist, line 38: There must be a cleaner way than using `w="100%"` no?

typescript/web/src/components/invitation-manager/invitation-container.tsx
```diff
@@ -0,0 +1,44 @@
...
+        p={6}
+        spacing={6}
+        align="start"
+        bg={mode("white", "gray.800")}
+        borderRadius="md"
+        boxShadow="lg"
+        maxW="2xl"
+        my="3.75rem"
+      >
+        <Heading fontSize="xl" fontWeight="semibold">
```
bbenoist, line 34: Why don't you use the as property with an appropriate value?

typescript/web/src/components/invitation-manager/invitation-container.tsx
bbenoist, line 21:
```suggestion
export type InvitationContainerProps = PropsWithChildren<{
  header: string;
  details: string;
}>;

export const InvitationContainer = ({
  header,
  details,
  children,
}: InvitationContainerProps) => {
```

typescript/web/src/components/invitation-manager/invitation-container.tsx
```diff
@@ -0,0 +1,44 @@
...
+          {header}
+        </Heading>
+        <Flex>{details}</Flex>
```
bbenoist, line 37:
```suggestion
        <Text>{details}</Text>
```

typescript/web/src/components/invitation-manager/invitation-container.tsx
```diff
@@ -0,0 +1,44 @@
+import {
+  Center,
+  Flex,
+  Heading,
```
bbenoist, line 4:
```suggestion
  Heading,
  Text,
```

bbenoist (changes requested):

typescript/web/src/components/invitation-manager/invitation-container.tsx
Jordanlelay: I should've explained it to you last time you made that comment.

`as` doesn't do anything on the style, it's just to include the right tag in the DOM. You can put whatever value you want in `as` it won't change the size.

typescript/web/src/components/invitation-manager/invitation-container.tsx
bbenoist: Apart from the fact that is ignores the padding and some other props, the parent is a flex container so using `width="100%"` is kind of an heresy. You should be able use `alignSelf="stretch"`  or something like that for such cases.

bbenoist (changes requested):
There is a defect when there is an error in the invitation. Please see my denied Chromatic capture for details.

Also, there remains the suggesion about `Flex width="100%"` within a `VStack` thing which I'd prefer to be fixed. It should just be the matter of switching to the correct prop...

typescript/web/src/components/invitation-manager/invitation-container.tsx
Jordanlelay: Not in all cases. If the text is small enough, the VStack will shrink
I've used it anyway and put a `w="full"` above it

Jordanlelay (commented):
@valerian could you give a look as @bbenoist and I both committed in this PR ?

bbenoist (approved):

Issue #860, opened 2022-02-07
Error when trying to import a COCO annotation file


nicOwlas:
I have following error when trying to import this V7 generated COCO annotation file to LabelFlow:

The COCO import never ends, the import appears as locked. If I open the dataset in another tab the import succeeded though.
- [Annotation File](https://github.com/labelflow/labelflow/files/8018599/lonzee-2022-02-05.zip)
- [Dataset to perform debug tests](https://labelflow.ai/labelflow/datasets/bug-import-coco)

![image](https://user-images.githubusercontent.com/16015322/152867616-ba47f1bd-b223-40c1-8480-686ad3445341.png)

nicOwlas:
@valerian : Link to the dataset on [drive](https://drive.google.com/drive/folders/15Dol-upAWz4IjtsBcv-r5zp1Q0UlUcL0?usp=sharing)

nicOwlas:
For now I would say ignore them silently and we'll add in the documentation that we only support polygons ie `iscrowd:0`

nicOwlas:
This could be the clear message. Could you add it?
2 tasks:

- [ ] Add a message with the number of ignored annotations if there are any.
- [ ] Change the icon for json files

![image](https://user-images.githubusercontent.com/16015322/153439009-de77dd3b-44b3-4494-95c3-a79f65d15304.png)

valerian:
> This could be the clear message. Could you add it?
> 2 tasks:

Ok, I will do that

PR #882, opened 2022-02-14
Fix coco import fail when containing crowd RLE annotations

valerian:
## Work performed

- Prevented COCO import from trowing an error when encountering crowd/RLE annotations, skipping them instead.
- On import modal:
  - Added a message with the number of ignored crowd/RLE annotations if there are any.
  - Changed the icon for JSON files and ZIP files.

## Resolved issues

#860

### DIFF
--- a/data/import.graphql
+++ b/data/import.graphql
@@ -1,5 +1,6 @@
 type ImportStatus{
     error: String
+    skippedCrowdAnnotations: Int!
 }
 
 input ImportOptionsCoco {
--- a/data/import.graphql
+++ b/data/import.graphql
@@ -1,6 +1,6 @@
 type ImportStatus{
     error: String
-    warnings: [String!]!
+    warnings: [String!]
 }
 
 input ImportOptionsCoco {
--- a/data/import.graphql
+++ b/data/import.graphql
@@ -1,6 +1,6 @@
 type ImportStatus{
     error: String
-    skippedCrowdAnnotations: Int!
+    warnings: [String!]!
 }
 
 input ImportOptionsCoco {
--- a/typescript/common-resolvers/src/import/format-coco/index.ts
+++ b/typescript/common-resolvers/src/import/format-coco/index.ts
@@ -216,11 +216,12 @@ async function importCocoAnnotationsIntoLabels(
   cocoImageIdToLabelFlowImageId: Map<number, string>,
   cocoCategoryIdToLabelFlowLabelClassId: Map<number, string>,
   { repository, req, user }: Context
-) {
+): Promise<{ skippedCrowdAnnotations: number }> {
   const indexedCocoImages = annotationFile.images.reduce(
     (imagesMap, image) => imagesMap.set(image.id, image),
     new Map<number, CocoImage>()
   );
+  let skippedCrowdAnnotations = 0;
   await Promise.all(
     annotationFile.annotations.map(async (annotation) => {
       if (!cocoImageIdToLabelFlowImageId.has(annotation.image_id)) {
@@ -228,7 +229,13 @@ async function importCocoAnnotationsIntoLabels(
           `Image ${annotation.image_id} referenced in annotation does not exist.`
         );
       }
-
+      if (annotation.iscrowd === 1) {
+        skippedCrowdAnnotations += 1;
+        console.log(
+          `Skipped annotation ${annotation.id}: crowd/RLE format no yet supported`
+        );
+        return;
+      }
       await labelResolvers.Mutation.createLabel(
         null,
         {
@@ -251,6 +258,7 @@ async function importCocoAnnotationsIntoLabels(
       console.log(`Created annotation ${annotation.id}`);
     })
   );
+  return { skippedCrowdAnnotations };
 }
 
 export const importCoco: ImportFunction = async (
@@ -276,10 +284,11 @@ export const importCoco: ImportFunction = async (
     cocoCategoryIdToLabelFlowLabelClassId,
     { repository, req, user }
   );
-  await importCocoAnnotationsIntoLabels(
+  const { skippedCrowdAnnotations } = await importCocoAnnotationsIntoLabels(
     annotationFile,
     cocoImageIdToLabelFlowImageId,
     cocoCategoryIdToLabelFlowLabelClassId,
     { repository, req, user }
   );
+  return { skippedCrowdAnnotations };
 };
--- a/typescript/common-resolvers/src/import/format-coco/index.ts
+++ b/typescript/common-resolvers/src/import/format-coco/index.ts
@@ -211,27 +211,35 @@ async function importCocoCategoriesIntoLabelClasses(
   }
 }
 
+function getSkippedCrowdAnnotationsWarning(
+  skippedCrowdAnnotations: number
+): string | undefined {
+  return skippedCrowdAnnotations > 0
+    ? `${skippedCrowdAnnotations} RLE bitmap annotations were ignored. Only polygon annotations are supported.`
+    : undefined;
+}
+
 async function importCocoAnnotationsIntoLabels(
   annotationFile: CocoDataset,
   cocoImageIdToLabelFlowImageId: Map<number, string>,
   cocoCategoryIdToLabelFlowLabelClassId: Map<number, string>,
   { repository, req, user }: Context
-): Promise<{ skippedCrowdAnnotations: number }> {
+): Promise<{ warning: string | undefined }> {
   const indexedCocoImages = annotationFile.images.reduce(
     (imagesMap, image) => imagesMap.set(image.id, image),
     new Map<number, CocoImage>()
   );
-  let skippedCrowdAnnotations = 0;
+  const annotationsToImport = annotationFile.annotations.filter(
+    (annotation) => annotation.iscrowd !== 1
+  );
+  const skippedCrowdAnnotations =
+    annotationFile.annotations.length - annotationsToImport.length;
   await Promise.all(
-    annotationFile.annotations.map(async (annotation) => {
+    annotationsToImport.map(async (annotation) => {
       if (!cocoImageIdToLabelFlowImageId.has(annotation.image_id)) {
         throw new Error(
           `Image ${annotation.image_id} referenced in annotation does not exist.`
         );
-      }
-      if (annotation.iscrowd === 1) {
-        skippedCrowdAnnotations += 1;
-        return;
       }
       await labelResolvers.Mutation.createLabel(
         null,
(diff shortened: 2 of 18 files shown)


typescript/web/src/components/import-button/import-images-modal/modal-dropzone/file-statuses.tsx
```diff
@@ -1,68 +1,131 @@
...
+        >
+          {isEmpty(droppedFile.errors) ? (
+            <ImportProgress status={fileUploadInfo?.status || false} />
+          ) : (
+            <ImportError errors={droppedFile.errors} />
+          )}
+        </Box>
+      </Flex>
+      {datasetSkippedCrowdAnnotations && (
+        <Box
```
bbenoist, line 86: image

typescript/web/src/components/import-button/import-images-modal/modal-dropzone/file-statuses.tsx
```diff
@@ -1,68 +1,131 @@
...
+          pr="2"
+          overflow="hidden"
+          textOverflow="ellipsis"
+          fontSize="xs"
+          color={mode("red.500", "red.300")}
+        >
+          {datasetSkippedCrowdAnnotations} RLE bitmap annotations were ignored.
+          Only polygon annotations are supported.
```
bbenoist, line 94: IMO, we don't have to display this. We can just mention that RLE annotations aren't supported in the documentation.
I'd would also remove the need to propagate the skippedCrowdAnnotations in a global status as I mentioned above.

typescript/common-resolvers/src/import/format-coco/index.ts
```diff
@@ -216,19 +216,26 @@ async function importCocoAnnotationsIntoLabels(
...
         throw new Error(
           `Image ${annotation.image_id} referenced in annotation does not exist.`
         );
       }
-
+      if (annotation.iscrowd === 1) {
+        skippedCrowdAnnotations += 1;
+        console.log(
+          `Skipped annotation ${annotation.id}: crowd/RLE format no yet supported`
+        );
```
bbenoist, line 236: Why do we need to display this in the backend console? Why is it not a warning?

bbenoist (changes requested):
As explained in the latest comment. I don't think we should display the skipped crowd annotations at all in the GUI.

typescript/web/src/components/import-button/import-images-modal/modal-dropzone/file-statuses.tsx
bbenoist: Ok so if we have to show this then we have to implement it in a more generic way (i.e. replace the skippedCrowd... value of ImportStatus by a warning array. Ok with that @valerian?

typescript/common-resolvers/src/import/format-coco/index.ts
valerian: Just keeping the same logic as in line 251/258 : `console.log(`Created annotation ${annotation.id}`);`
But yeah we probably shouldn't have any of them permanently in the backend logs ...

typescript/web/src/components/import-button/import-images-modal/modal-dropzone/file-statuses.tsx
valerian: The only other location being [typescript/web/src/components/website/Pricing/CardBadge.tsx:19](https://github.com/labelflow/labelflow/blob/bug/coco-import-error-crowd-rle/typescript/web/src/components/website/Pricing/CardBadge.tsx#L19), I don't think factorizing would make much sense

typescript/common-resolvers/src/import/format-coco/index.ts
```diff
@@ -276,10 +281,18 @@ export const importCoco: ImportFunction = async (
...
     annotationFile,
     cocoImageIdToLabelFlowImageId,
     cocoCategoryIdToLabelFlowLabelClassId,
     { repository, req, user }
   );
+  return {
+    warnings:
+      skippedCrowdAnnotations > 0
+        ? [
+            `${skippedCrowdAnnotations} RLE bitmap annotations were ignored.\nOnly polygon annotations are supported.`,
```
bbenoist, line 294:
```suggestion
            `${skippedCrowdAnnotations} RLE bitmap annotations were ignored. Only polygon annotations are supported.`,
```

data/import.graphql
```diff
@@ -1,5 +1,6 @@
 type ImportStatus{
     error: String
+    warnings: [String!]!
```
bbenoist, line 3: Can we make the output nullable so that it's consistent with error and more "standard-ish"?

typescript/web/src/components/import-button/import-images-modal/modal-dropzone/file-statuses.tsx
```diff
@@ -1,68 +1,130 @@
...
+
+const FileStatus = ({
+  droppedFile,
+  index,
+  fileUploadInfo,
+}: {
+  droppedFile: DroppedFile;
+  index: number;
+  fileUploadInfo?: UploadInfo;
+}) => {
```
bbenoist, line 43: Please move the type definition into into its own `FileStatusProps` type

typescript/web/src/components/import-button/import-images-modal/modal-dropzone/file-statuses.tsx
```diff
@@ -1,68 +1,130 @@
...
+  const warnings = fileUploadInfo?.warnings;
+  return (
+    <>
```
bbenoist, line 46: This block is becoming very big. Can you split it into smaller parts?

typescript/web/src/components/import-button/import-images-modal/types.ts
```diff
@@ -24,7 +24,14 @@ export type DroppedUrl = {
  */
 export type UploadStatuses = Record<string, boolean | string>;
 
+export type UploadInfo = {
+  status: boolean | string;
```
bbenoist, line 28: Why do you need to use boolean | string? Isn't `?: string` sufficient?

bbenoist (changes requested):

typescript/web/src/components/import-button/import-images-modal/types.ts
```diff
@@ -24,7 +24,14 @@ export type DroppedUrl = {
...
+  warnings: string[];
+};
+export type UploadInfos = Record<string, UploadInfo>;
```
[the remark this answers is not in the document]
valerian: I can't rename `UploadInfos` to `UploadsInfo` because `UploadsInfo` already exists, maybe I can use `UploadInfoArray` ?

typescript/web/src/components/import-button/import-images-modal/types.ts
valerian: it has 3 states: not uploaded (false), uploaded (true), or error (string).

typescript/web/src/components/import-button/import-images-modal/modal-dropzone/file-statuses.tsx
```diff
@@ -1,68 +1,147 @@
...
+    overflow="hidden"
+    textOverflow="ellipsis"
+    whiteSpace="nowrap"
+  >
+    {droppedFile.file.path}
+  </Box>
+);
+
+const Warnings = ({ warnings }: { warnings: string[] | undefined }) => (
+  <>
```
bbenoist, line 68: No VStack, ul or table here?

typescript/web/src/components/import-button/import-images-modal/modal-dropzone/file-statuses.tsx
```diff
@@ -1,68 +1,147 @@
...
-              <ImportProgress
-                status={fileUploadStatuses[file.path ?? file.name]}
-              />
-            ) : (
-              <ImportError errors={errors} />
-            )}
-          </Box>
-        </Flex>
+      {files.map((droppedFile, index) => (
+        <FileStatus
```
bbenoist, line 138: We need a key here too (⚠️ [_using index will not be possible_](https://medium.com/geekculture/reactjs-why-index-as-a-key-is-an-anti-pattern-4b9dc6ef0067)).

typescript/web/src/components/import-button/import-images-modal/types.ts
bbenoist: OK thanks for the answer :) Then can we do a **_quick-change_** to use this instead?

```ts
status: FileUploadStatus
error?: string;
```

```ts
type FileUploadStatus = "pending" | "uploading" | "uploaded" | "error";
```

**_Side-note:_** _If we started from scratch, we could have used a [react-query](https://react-query.tanstack.com/overview) mutation which mimicks the Apollo `useMutation` output ( with the `data`, `error` and `loading` names etc.), hence simplifying and standardizing the interface we're talking about 🚀_  
_But changes would have been too big for this PR anyway, it's just some knowledge-sharing 😉 You've done the right job!_

bbenoist (changes requested):

typescript/web/src/components/import-button/import-images-modal/types.ts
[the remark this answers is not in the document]
bbenoist: Sorry, I can't explain why but this was an incorrect version of my comment 🤦‍♂️ The `Rename the UploadInfos type into UploadsInfo.` line should have been replaced by `Remove the UploadInfos type as it's only used once.`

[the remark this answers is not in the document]
valerian: well, it's used a lot in `typescript/web/src/components/import-button/import-images-modal/modal-dropzone/file-statuses.tsx`

typescript/web/src/components/import-button/import-images-modal/modal-dropzone/file-statuses.tsx
valerian: there is a `Flex` component in `FileStatus` in which those will go

typescript/web/src/components/import-button/import-images-modal/types.ts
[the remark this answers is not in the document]
bbenoist: It's used as a variable and inferred but from what I see in the changes tab, the UploadInfoRecord type is only explicitly mentioned once.

typescript/web/src/components/import-button/import-images-modal/modal-dropzone/file-statuses.tsx
bbenoist: OK fine 👍 Thanks for the answer!

bbenoist (approved):

Issue #906, opened 2022-02-21
Fix CSV export


nicOwlas:
- [x] White icons
- [x] Add tests
- [x] Remove option to add URLs (always have it in the CSV)
- [x] ~~Set the validity period to: 30 days~~

valerian:
@nicOwlas what to display when we don't have options? 🤔
![Screenshot 2022-02-21 at 16 20 35](https://user-images.githubusercontent.com/239688/154983607-c61627e0-0933-43b2-b7b8-ce206cc5c53c.png)

nicOwlas:
Can you skip the option modal and directly launch the download of the CSV?

valerian:
AWS can't create signed URLs that expire in more than 7 days:
https://aws.amazon.com/premiumsupport/knowledge-center/presigned-url-s3-bucket-expiration/

nicOwlas:
@valerian we should be putting the `labelId` in the first column instead of the `imageId` no?

PR #908, opened 2022-02-21
Fix CSV export

valerian:
## Resolved issues

Solves #906


typescript/web/src/components/export-button/export-modal/export-options-modal.tsx
```diff
@@ -100,12 +103,17 @@ export const ExportOptionsModal = () => {
...
-  const handleClick = useCallback(() => {
+  const exportAction = useCallback(() => {
     exportFunction(exportOptions);
     trackEvent(`export_button_click_${exportFormat.toLocaleLowerCase()}`, {});
     setIsOptionsModalOpen(false);
   }, [exportFormat, exportFunction, exportOptions, setIsOptionsModalOpen]);
 
+  useEffect(() => {
+    if (isOptionsModalOpen && formatOptionsInformationArray.length === 0)
+      exportAction();
```
bbenoist, line 114:
```suggestion
    if (isOptionsModalOpen && formatOptionsInformationArray.length === 0) {
      exportAction();
    }
```

typescript/web/src/components/export-button/export-modal/export-options-modal.tsx
```diff
@@ -100,12 +103,17 @@ export const ExportOptionsModal = () => {
...
   );
 
-  const handleClick = useCallback(() => {
+  const exportAction = useCallback(() => {
     exportFunction(exportOptions);
     trackEvent(`export_button_click_${exportFormat.toLocaleLowerCase()}`, {});
     setIsOptionsModalOpen(false);
   }, [exportFormat, exportFunction, exportOptions, setIsOptionsModalOpen]);
 
+  useEffect(() => {
```
bbenoist, line 112: Why do you need a `useEffect` exactly? Can't it be avoided?

typescript/web/src/components/export-button/export-modal/export-card.tsx
```diff
@@ -92,7 +94,17 @@ export const ExportCard = ({
...
         >
-          <Image src={logoSrc} w="16" flexGrow={0} flexShrink={0} />
+          <Image
+            src={logoSrc}
+            w="16"
+            flexGrow={0}
+            flexShrink={0}
+            filter={
+              logoUsesColorMode
+                ? mode("brightness(0)", "brightness(0) invert()")
```
bbenoist, line 104: When importing the SVG directly from the react files instead of relying on `/static`, you should be able to change colors of the logo without having to hack brightness

data/export.graphql
```diff
@@ -19,7 +19,6 @@ input ExportOptionsYolo {
 
 input ExportOptionsCsv {
   name: String
-  includeImageUrl: Boolean
 }
```
bbenoist, line 22: Why 😿 ?

valerian: As discussed on Monday's DSM, it should be by default, and it will also be by default on other export formats soon (unless with pick the option to include the whole images in the export)

typescript/web/src/components/export-button/export-modal/export-options-modal.tsx
valerian: I'm not sure, but if I don't use it I get this in the console when clicking on the CSV export card:

```
Warning: Cannot update a component (`ExportModalProvider`) while rendering a different component (`ExportOptionsModal`).
```

I'm no expert in React hooks, I would appreciate your help on that.

typescript/common-resolvers/src/export/export-to-csv.fixtures.ts
```diff
@@ -0,0 +1,147 @@
...
+    url: "https://localhost:3000/api/downloads/8f47e891-3b24-427a-8db0-dab362fbe269/9b8fb142-8388-11ec-a8ee-f7dcb0508f86.jpg/original.jpeg",
+    path: "https://localhost:3000/api/downloads/8f47e891-3b24-427a-8db0-dab362fbe269/9b8fb142-8388-11ec-a8ee-f7dcb0508f86.jpg/original.jpeg",
+    thumbnail200Url:
+      "https://localhost:3000/api/downloads/8f47e891-3b24-427a-8db0-dab362fbe269/9b8fb142-8388-11ec-a8ee-f7dcb0508f86.jpg/t200.jpeg",
+    mimetype: "image/jpeg",
+    width: 1920,
+    height: 1080,
+  },
+  {
+    id: "a552037e-8388-11ec-965b-07327355c580",
```
bbenoist, line 32: Can you merge both declarations in a map? 
image

typescript/common-resolvers/src/export/export-to-csv.fixtures.ts
```diff
@@ -0,0 +1,147 @@
...
+      return LABEL_CLASSES_DATA;
+    },
+  },
+  label: {
+    list: async (where) => {
+      if (where?.datasetId !== DATASET_DATA.id)
+        throw new Error("Unexpected parameters");
+      return LABELS_DATA;
+    },
+  },
```
bbenoist, line 146: The same function is copy-pasted 4 times here. Can you factorize it a little bit 🙏 ?

typescript/web/src/components/export-button/export-modal/export-card.tsx
```diff
@@ -92,7 +97,16 @@ export const ExportCard = ({
...
+          {LogoIcon ? (
+            <LogoIcon
+              fontSize="64"
+              color={mode("black", "white")}
+              flexGrow={0}
+              flexShrink={0}
+            />
+          ) : (
+            <Image src={logoSrc} w="16" flexGrow={0} flexShrink={0} />
+          )}
```
bbenoist, line 109: Can you move this block into an ExportIcon component so that we don't increare ExportCard complexity?

typescript/web/src/components/export-button/export-modal/formats.ts
```diff
@@ -14,16 +16,24 @@ export const DEFAULT_EXPORT_OPTIONS: RequiredExportOptions = {
...
+    logoSrc?: string;
+    logoIcon?: IconType;
+  } & (
+    | {
+        logoSrc: string;
+      }
+    | {
+        logoIcon: IconType;
+      }
+  );
```
bbenoist, line 36: Can we move the type into its own declaration?

bbenoist (changes requested):
Except for the last small factorization which should not take more than 10mn, it's all good now 👏

PR #909, opened 2022-02-22
Make thumbnails appear or disappear when adding or deleting images

### LINKED ISSUE #898: Thumbnails are not showing up in labeling view
### Bug 1
- Open a dataset having already some images uploaded
- Open the labeling view
- Add new images
- The new images thumbnails are not showing in the gallery unless one needs to refresh

After a refresh the image thumbnails can been seen.
### Bug 2
- Open the image gallery view of the previous dataset
- Delete some images
- Open the labeling view
- The thumbnails of the previous images can be still be seen though they were deleted

After a refresh the image thumbnails are not visible anymore.

## Screen recording

https://user-images.githubusercontent.com/16015322/154306177-c438bd79-cb9d-42a5-931c-966a661647f6.mov

Jordanlelay:
## Work performed

- Update refetch queries when creating an image
- Query all images of a dataset when deleting an image to update navigation images

## Resolved issues

Closes #898

### DIFF
--- a/typescript/web/src/components/dataset-images-list/delete-image-modal.tsx
+++ b/typescript/web/src/components/dataset-images-list/delete-image-modal.tsx
@@ -1,5 +1,4 @@
-import { gql, useQuery, useMutation, useApolloClient } from "@apollo/client";
-import { useCallback, useRef } from "react";
+import { gql, useMutation, useQuery } from "@apollo/client";
 import {
   AlertDialog,
   AlertDialogBody,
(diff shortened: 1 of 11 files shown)


typescript/web/src/components/dataset-images-list/delete-image-modal.tsx
```diff
@@ -37,6 +43,23 @@ const DELETE_IMAGE_MUTATION = gql`
...
+  const { slug: datasetSlug } = useDataset();
+  return useCallback(async () => {
+    client.query<
+      GetAllImagesOfADatasetQuery,
+      GetAllImagesOfADatasetQueryVariables
+    >({
+      query: GET_ALL_IMAGES_OF_A_DATASET_QUERY,
+      variables: { slug: datasetSlug, workspaceSlug },
+      fetchPolicy: "network-only",
+    });
```
bbenoist, line 59: Why don't you use `refetchQueries`?

bbenoist (changes requested):

typescript/web/src/components/dataset-images-list/delete-image-modal.tsx
Jordanlelay: Because the query does not exist here.

bbenoist: So why do we need to fetch it at all?

Jordanlelay: To refresh the cache, which contains the wrong data. The query is not active but the data remains, so we need to refetch everything in order to have the correct data displayed.

bbenoist:
Can't we just clear it instead then?

Baptist
________________________________

Jordanlelay:
> Can't we just clear it instead then?

What would be the difference in the end ? Both methods will invalidate the cache and launch the query as I did, but the way I did it avoids to manage manually cache entries which, in my opinion, won't be beneficial.

valerian (approved):

bbenoist (approved):

PR #913, opened 2022-02-22
Support trial period in pricing components and store Stripe subscription status

### LINKED ISSUE #531: Manage trial period
- [x] Edit website for starter / pro plan call to actions: Change `Request Access` for `Start Free trial`
- [x] Write below the Starter and Pro plan that the trial is 14 days
- [x] Redirect to the sign-in page then on the new workspace page when clicking `Start 14 day trial`. Create the workspace with the plan corresponding to the one selected by the user.: https://labelflow.ai/local/datasets?modal-create-workspace&workspace-name= 
- [x] After 2 weeks: display a warning banner: See mockup for the text.

Stripe triggers an event when a workspace subscription changes.
- [x] Add a status field in the workspace entity graphql to know if a plan has expired (copy stripe statuses)
- [x] Listen to Stripe webhook to update the workspace status (has expired or not)`typescript/web/src/pages/api/stripe/webhooks.ts`
- [x] Make sure there is an error if the plan has expired (in order to display the alert banner)

## Not part of this issue
```
- [ ] Disabled features when trial has expired
  - [ ] Only the first workspace remains active. Other workspaces are inactive. Only the settings page remain accessible.
  - [ ] In the active workspace: only the datasets summing to a max of 1000 images remain fully active (possibility to add images to the limit of 1000 and possibility to export)
  - [ ] The other datasets are read only, export is disabled.
```

### Banner ([here](https://www.figma.com/file/4J9xyfF5v4B98MuKVO0Y8V/LabelFlow-mockups?node-id=4505%3A17853))
image

### Pricing ([here](https://www.figma.com/file/4J9xyfF5v4B98MuKVO0Y8V/LabelFlow-mockups?node-id=4543%3A18625))
image

Blocked by #799

Jordanlelay:
## Work performed

- [x] Add a banner on datasets page if workspace status is not valid
- [x] Add a 14 days trial period
- [x] Add a status field in workspace entity
- [x] Make the workspace plan field variable (previously always `community`
- [x] Update tests
- [x] get status from Stripe after the subscription created

Closes #531


typescript/web/src/components/workspaces/update-plan-banner.tsx
```diff
@@ -0,0 +1,38 @@
+import { Button, chakra, Flex, Text } from "@chakra-ui/react";
+import { useRouter } from "next/router";
+import { RiInformationFill } from "react-icons/ri";
+import { useWorkspace } from "../../hooks/use-user";
+
+const InfoIcon = chakra(RiInformationFill);
+export const UpdatePlanBanner = () => {
```
bbenoist, line 7: image

IDK if we talked about it yet in DSM or other but you can install the CodeClimate Chrome extension to see such lines in GitHub diff 😉

bbenoist, line 7: image

IDK if we talked about it yet in DSM or other but you can install the CodeClimate Chrome extension to see such lines in GitHub diff 😉

bbenoist, line 7: image

IDK if we talked about it yet in DSM or other but you can install the CodeClimate Chrome extension to see such lines in GitHub diff 😉

typescript/web/src/components/workspaces/update-plan-banner.tsx
```diff
@@ -0,0 +1,38 @@
...
+  const redirectToSettings = () => {
+    router.push(`/${workspaceSlug}/settings`);
+  };
+  return (
+    <Flex
+      direction={{ base: "column", md: "row" }}
+      justify="center"
+      align="center"
+      bg="#EFAB22"
+      zIndex="100"
```
bbenoist, line 19: You shouldn't have to use zIndex but place the banner at the correct place in the DOM tree

typescript/web/src/components/workspaces/update-plan-banner.tsx
```diff
@@ -0,0 +1,38 @@
...
+      p="4"
+      gridGap="4"
+    >
+      <InfoIcon h="24px" w="24px" />
+      <Text fontWeight="500" textAlign="center">
+        Your trial has expired. Update your payment info to not lose access to
+        your datasets or downgrade to the community plan.
```
bbenoist, line 26: @nicOwlas I'm not that an expert in English but isn't there a grammar problem with `to not lose access to`? Shouldn't it be `to don't lose access to`?

Anyway what about `to keep access to`

typescript/web/src/components/workspaces/update-plan-banner.tsx
```diff
@@ -0,0 +1,38 @@
...
+      </Text>
+      <Button
+        bg="orange.400"
+        size="sm"
+        _hover={{ bg: "orange.200" }}
+        onClick={redirectToSettings}
+      >
+        <Text>Update Plan</Text>
```
bbenoist, line 34: @nicOwlas Shouldn't we use `Upgrade plan`?

bbenoist, line 34: @Jordanlelay You don't need to wrap your button text with `Text` components. https://chakra-ui.com/docs/form/button#usage

typescript/web/src/pages/[workspaceSlug]/datasets/index.tsx
```diff
@@ -12,12 +12,19 @@ import { WorkspaceTabBar } from "../../../components/layout/tab-bar/workspace-ta
...
 import { BoolParam, IdParam } from "../../../utils/query-param-bool";
 
 const Body = () => {
-  const { slug: workspaceSlug } = useWorkspace();
+  const { slug: workspaceSlug, status } = useWorkspace();
   const { isReady } = useRouter();
+  const isWorkspaceActive =
+    status === WorkspaceStatus.active ||
+    status === WorkspaceStatus.trialing ||
+    status === WorkspaceStatus.incomplete;
```
Comment (bbenoist) [inline typescript/web/src/pages/[workspaceSlug]/datasets/index.tsx:26]: While we're at introducting a @labelflow/core package, I'd move this logic into a shared `isWorkspaceActive(status: WorkspaceStatus)` function. Or at least in `@labelflow/common-resolvers` if you don't want to follow the _"little-bit-less quick"_ solution I've detailed above.

typescript/web/src/pages/[workspaceSlug]/datasets/index.tsx
```diff
@@ -55,6 +62,7 @@ const Body = () => {
 
   return (
     <>
+      {shouldDisplayUpdateBanner && <UpdatePlanBanner />}
```
Comment (bbenoist) [inline typescript/web/src/pages/[workspaceSlug]/datasets/index.tsx:65]: As it'll possibly be reused in many pages, the boolean logic should be placed within UpdatePlanBanner which will take care of the responsibility to add the necessary components or not.

bbenoist (changes requested):

typescript/web/src/components/workspaces/update-plan-banner.tsx
nicOwlas: `To not lose access to` is grammatically correct but I agree that the double negation should be avoided. Ok to replace with `to keep access to`

nicOwlas, line 26:
```suggestion
        Your trial has expired. Update your payment info to keep access to
        your datasets or downgrade to the community plan.
```

typescript/web/src/components/workspaces/update-plan-banner.tsx
nicOwlas: It's not really an upgrade. We want the user to update his payment method etc.

typescript/db/src/utils/stripe.ts
```diff
@@ -1,6 +1,23 @@
+import { WorkspacePlan } from "@labelflow/graphql-types";
+import { DAYS_OF_TRIAL, DEFAULT_WORKSPACE_PLAN } from "@labelflow/utils";
+import { getUnixTime, addDays } from "date-fns";
 import { isNil } from "lodash/fp";
 import Stripe from "stripe";
 
+const planToStripePriceId: Record<WorkspacePlan, string> = {
+  Community: process.env.STRIPE_COMMUNITY_PLAN_PRICE_ID ?? "",
+  Starter: process.env.STRIPE_STARTER_PLAN_PRICE_ID ?? "",
+  Pro: process.env.STRIPE_PRO_PLAN_PRICE_ID ?? "",
```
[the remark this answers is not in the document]
Jordanlelay: Something in that sense is done by the function `stripeIsDefined` line 21.

Also, why would we want to do that test ? It opens the possibility of having none of the variables defined. It looks safer to check that all variables are defined, which is already done from lines 30 -> 38

data/entities.graphql
```diff
@@ -126,11 +127,22 @@ enum WorkspacePlan {
...
+enum WorkspaceStatus {
+  active
+  trialing
+  incomplete
+  past_due
+  unpaid
+  canceled
+  incomplete_expired
+  ended
+  all
```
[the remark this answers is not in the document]
Jordanlelay: It's a mistake, I'll remove it

typescript/web/src/pages/[workspaceSlug]/datasets/index.tsx
Comment (Jordanlelay) [reply typescript/web/src/pages/[workspaceSlug]/datasets/index.tsx:26]: Why not leave it here as it's not yet used anywhere else ?

typescript/web/src/pages/[workspaceSlug]/datasets/index.tsx
Comment (Jordanlelay) [reply typescript/web/src/pages/[workspaceSlug]/datasets/index.tsx:65]: Possibly.
But for now, there is an advantage to use this way: you don't pollute virtual DOM with useless components
Capture d’écran 2022-02-24 à 17 54 31

typescript/web/src/components/workspaces/update-plan-banner.tsx
Jordanlelay: Based on the discussion we had with @nicOwlas today, shouldn't we leave code metrics to the side for the moment as it's not that critical right now ?

typescript/db/src/utils/tests/create-workspace.mutation.ts
```diff
@@ -35,7 +36,11 @@ export const createWorkspace = (
   }>({
     mutation: CREATE_WORKSPACE_MUTATION,
     variables: {
-      data: { ...data, name: data?.name === undefined ? "test" : data.name },
+      data: {
+        ...data,
+        name: data?.name === undefined ? "test" : data.name,
+        plan: data?.plan === undefined ? WorkspacePlan.Pro : data.plan,
```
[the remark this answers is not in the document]
Jordanlelay: Yes but for some reason I have to set it here as well, otherwise tests are failing. As it is a test mutation, I propose to leave it like that

typescript/web/src/pages/[workspaceSlug]/datasets/index.tsx
Comment (bbenoist) [reply typescript/web/src/pages/[workspaceSlug]/datasets/index.tsx]: It will not pollute the DOM. I've just asked for something like:

```tsx
const UpdatePlanBanner = () => {
  const showBanner = // get the info
  return <>{showBanner && <UpdatePlanBannerComponent /></>
}
```

There is no extra component added with such code. The `<>` fragment does not creates any DOM component,

Comment (Jordanlelay) [reply typescript/web/src/pages/[workspaceSlug]/datasets/index.tsx]: I've already updated it

typescript/db/src/utils/stripe.ts
```diff
@@ -1,6 +1,25 @@
...
+
+const getPriceId = (
+  metadata: Stripe.Emptyable<Stripe.MetadataParam> | undefined
+): string => {
+  if (metadata && !isNil(metadata.plan)) {
+    return planToStripePriceId[metadata.plan as WorkspacePlan];
+  }
+  return planToStripePriceId[DEFAULT_WORKSPACE_PLAN];
+};
 export function stripeIsDefined(
```
bbenoist, line 23:
```suggestion

export function stripeIsDefined(
```

typescript/db/src/utils/stripe.ts
```diff
@@ -1,6 +1,25 @@
...
+  Community: process.env.STRIPE_COMMUNITY_PLAN_PRICE_ID ?? "",
+  Starter: process.env.STRIPE_STARTER_PLAN_PRICE_ID ?? "",
+  Pro: process.env.STRIPE_PRO_PLAN_PRICE_ID ?? "",
+};
+
+const getPriceId = (
+  metadata: Stripe.Emptyable<Stripe.MetadataParam> | undefined
+): string => {
+  if (metadata && !isNil(metadata.plan)) {
+    return planToStripePriceId[metadata.plan as WorkspacePlan];
```
bbenoist, line 19: Before casting, it's necessary to check whether `metadata.plan` matches one of the possible values of `WorkspacePlan`.

typescript/db/src/utils/tests/create-workspace.mutation.ts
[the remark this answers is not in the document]
bbenoist: Yes but can you at-least reuse DEFAULT_WORKSPACE_PLAN since it what should be returned by the server?

typescript/web/src/components/workspace-switcher/create-workspace-modal/create-workspace-modal.tsx
```diff
@@ -101,8 +103,13 @@ export const useCreateWorkspace = (): [
   string | undefined
 ] => {
   const { name } = useWorkspaceNameInput();
+  const [plan] = useQueryParam("plan", StringParam);
+  const nonUndefinedPlan = plan ?? "";
+  const planToCreate = !(nonUndefinedPlan in WorkspacePlan)
+    ? DEFAULT_WORKSPACE_PLAN
```
bbenoist, line 109: We should throw an error if the workspacePlan given in query params doesn't matches our own list.

typescript/web/src/components/workspaces/update-plan-banner.stories.tsx
```diff
@@ -0,0 +1,16 @@
+import { createCommonDecorator, storybookTitle } from "../../utils/stories";
+import { WORKSPACE_DATA } from "../../utils/fixtures";
+import { UpdatePlanBanner } from "./update-plan-banner";
+
+export default {
+  title: storybookTitle("Workspaces", "UpdatePlanBanner"),
```
bbenoist, line 6:
```suggestion
  title: storybookTitle("Workspaces", UpdatePlanBanner),
```

typescript/web/src/components/workspaces/update-plan-banner.stories.tsx
```diff
@@ -0,0 +1,16 @@
...
+  decorators: [
+    createCommonDecorator({
+      auth: { withWorkspaces: true },
+      apollo: true,
+      router: { query: { workspaceSlug: WORKSPACE_DATA.slug } },
+    }),
+  ],
+};
+
+export const Open = () => <UpdatePlanBanner />;
```
bbenoist, line 16: It's kind of an hidden Storybook feature but if you name your single story with the same name as the component it exposes, you will remove the useless extra-level in the left-sidebar 😉 

Also, we should give the whole component as last argument to `storybookTitle` instead of just it's name so that it not more portable.

```suggestion
import { UpdatePlanBanner as UpdatePlanBannerComponent } from "./update-plan-banner";

export default {
  title: storybookTitle("Workspaces", UpdatePlanBannerComponent),
  component: UpdatePlanBannerComponent,
  decorators: [
    createCommonDecorator({
      auth: { withWorkspaces: true },
      apollo: true,
      router: { query: { workspaceSlug: WORKSPACE_DATA.slug } },
    }),
  ],
};

export const UpdatePlanBanner = () => <UpdatePlanBannerComponent />;
```

bbenoist (changes requested):

typescript/web/src/components/website/Pricing/Pricing.tsx
```diff
@@ -48,17 +42,22 @@ export const Pricing = () => (
...
         }}
         icon={BsPeopleFill}
         button={
-          <NextLink href="/auth/signin">
+          <NextLink
+            href={{
+              pathname: "/auth/signin",
+              query: { redirect: "/workspaces?plan=community" },
+            }}
+          >
```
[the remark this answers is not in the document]
bbenoist: We don't need to pass-by `/auth/signin`, the redirection will be operated automatically by `/workspaces` as for all of our webpages which requires an authentication.
And my suggestion does works. The href property of `next/link` component accepts a string. That's what's used everywhere in the code:
image

typescript/web/src/components/workspaces/update-plan-banner.tsx
bbenoist: Have I already blocked a PR just because of CodeClimate feedbacks?
No because it's already on the side, It does not blocks the CI and we don't consider the warnings which involves too much time to be fixed.

One of the point of using CodeClimate is to help us getting code-quality feedbacks quicker than human code-reviews. It does not replaces them but avoids the need to manually tell you the obvious code smells.

typescript/web/src/components/workspaces/update-plan-banner.tsx
```diff
@@ -0,0 +1,45 @@
+import { Button, chakra, Flex, Text } from "@chakra-ui/react";
+import { WorkspaceStatus } from "@labelflow/graphql-types";
+import { useRouter } from "next/router";
+import { RiInformationFill } from "react-icons/ri";
+import { useWorkspace } from "../../hooks/use-user";
+
+const InfoIcon = chakra(RiInformationFill);
+export const UpdatePlanBanner = () => {
```
bbenoist, line 8:
```suggestion
export const UpgradePlanBanner = () => {
```

(don't forget to rename the file and its story too 😉 )

typescript/web/src/pages/api/stripe/webhooks.ts
```diff
@@ -87,6 +91,12 @@ const webhookHandler = async (req: NextApiRequest, res: NextApiResponse) => {
           throw new Error("No workspace found for this subscription");
         }
 
+        if (!(status in WorkspaceStatus)) {
```
bbenoist, line 94: By using `in`, you're comparing `status` with the keys of `WorkspaceStatus`, not its values. You have to compare with the values of `WorkspaceStatus`.

typescript/web/src/pages/api/stripe/webhooks.ts
```diff
@@ -87,6 +91,12 @@ const webhookHandler = async (req: NextApiRequest, res: NextApiResponse) => {
...
+          throw new Error(`Unknown status ${status}`);
+        }
+
+        const workspaceStatus =
+          WorkspaceStatus[pascalCase(status) as keyof typeof WorkspaceStatus];
```
bbenoist, line 99:
```suggestion
        const workspaceStatus = pascalCase(status) as WorkspaceStatus;
```

data/entities.graphql
```diff
@@ -126,11 +127,20 @@ enum WorkspacePlan {
...
-  Enterprise
 }
 
 enum MembershipRole {
   Owner
   Admin
   Member
 }
+
+enum WorkspaceStatus {
```
bbenoist, line 138:
```suggestion
// https://stripe.com/docs/api/subscriptions/object#subscription_object-status
enum WorkspaceStatus {
```

typescript/web/src/pages/[workspaceSlug]/datasets/index.tsx
Comment (bbenoist) [reply typescript/web/src/pages/[workspaceSlug]/datasets/index.tsx]: Because we'll need is as-soon as we'll have to block features if a workspace is inactive.

typescript/web/src/pages/api/stripe/webhooks.ts
Jordanlelay: No. I'm comparing `status` against the keys for a good reason, as I'm doing `WorkspaceStatus[status]` after.
Why would I compare if the `status` is in the values of my enum ?

data/entities.graphql
Jordanlelay: comments in graphql start with `#`

bbenoist: OK. Thanks for having done the correct change instead 👍

typescript/web/src/pages/api/stripe/webhooks.ts
bbenoist: Yes, you've used the key below, that's why I've added [a suggestion](https://github.com/labelflow/labelflow/pull/913/files#r816625316) on the corresponding line.

Except for very specific cases, with enums we store the value and compare against it, not the key. The key is just an alias to the value it's not guaranteed that they are equal (_spoiler, they can be different if you change the configuration of graphql-codegen or Prisma, which is what I had to do at some point in #685 for example_). We already have some parts in the code with different keys and values (e.g. `Tools`).

You can see TypeScript enums as a way to reference string or int values with constants instead of plain values, It's actually even just transpiled as a JS const object.
i.e. instead of the plain `"Active"` word if `WorkspaceStatus` was not an enum but a union like `"Active" | "Trialing" | "Incomplete" | ...`, you can use `WorkspaceStatus.Active`.
But it's not more than that, the key is not expected to be exposed outside of the code, that's the purpose of the value.

You could eventually support such requirements but it'd need some very strong guards making sure that the enum values matches the keys (since assuming that they are equals without it would break the _single-responsibility principle_). But at the end, it's just more complex than relying on the value.

bbenoist (commented):
@Jordanlelay I've pushed a batch of cleanups + the useEnumValue changes + the changes we've talked about. You can look at the individuals commits for details. Let me know in comments if you think that some changes has to be done and I'll revert the corresponding commit or do the changes quickly.

Other than that, we have to take care of the migration. As said during the DSM, we can mob it tomorrow if you're OK.

nicOwlas:
I tried it on the deployment branch and I can't see the usage bar in the dev DB workspaces @Jordanlelay. Any idea why?
image

Jordanlelay:
> I tried it on the deployment branch and I can't see the usage bar in the dev DB workspaces @Jordanlelay. Any idea why?

Last commits come from @bbenoist but I can give a look

Jordanlelay:
@nicOwlas It works for me with a local build with both a local database and the dev database. Something must be wrong with the deployment. By the way, given the screenshot you provided, it looks like you use an outdated preview

typescript/web/src/pages/auth/signin.tsx
```diff
@@ -22,7 +22,9 @@ const useRedirectIfAuthenticated = (redirectUrl?: string) => {
 };
 
 const MustLogInMessage = () => (
-  <Heading as="h3">Please sign-in to continue</Heading>
+  <Heading as="h3" textAlign="center" fontSize="3xl">
+    Please sign-in to continue
+  </Heading>
```
bbenoist, line 27: Thank you for this fix ❤️  It's sad that I didn't moved the whole sign-in page body in a component when I've built the page. We would have it covered with Chromatic if it was the case...

_You can self-resolve this comment once read_

typescript/db/schema.prisma
```diff
@@ -95,15 +95,16 @@ model Dataset {
...
+  updatedAt        DateTime        @updatedAt
   deletedAt        DateTime?
-  name             String        @unique
-  slug             String        @unique
+  name             String          @unique
+  slug             String          @unique
   image            String?
   plan             WorkspacePlan
   stripeCustomerId String?
+  status           WorkspaceStatus @default(Active)
```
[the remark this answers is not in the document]
Jordanlelay: Well, I've put this value here because of the existing workspaces in db (and it's required if I remember the error message correctly), which all are `Community` ones. I don't see why we should put those workspaces in trialing mode

typescript/web/src/pages/auth/signin.tsx
Jordanlelay: I'm not even sure that this message should be here, as it's not in the Figma mockups. I've pushed this change to re-trigger the whole CI as it was failing for whatever reason

bbenoist: For the location, it feel non optimal I agree. As it's not the goal of this PR we can keep it as-is and we'll eventually open an issue during today's dsm if needed 😉

bbenoist (approved):
I've added 2 new commits, feel free to discuss the changes with me if there is a problem somewhere:
* With your last commit, the status was not converted to PascalCase before casting it to `WorkspaceStatus`. I've fixed it by calling `pascalCase` from _change-case_ and by using the `toEnumValue` function introduced in this PR.
* I've moved the workspace image count query from the workspace resolver to the workspace repository and exposed the result as a `number` so that the existing `imageAggregates` query can just  call `repository.workspace.countImages`.

Things are now ready to be merged IMO ✅ 

Pareto's law got the feature come later than expected as we found a few edge-cases and there was the "enum problem". But you've done a good work with the 2 PR we're about to merge!

* You had a good understanding of Stripe. Except from the "status at creation" edge-case, the process was very well handled and you were very autonomous and perseverant 👏 _Kudos for having found the status in the "not that clear" Stripe TS types 🥇_

* The frontend design looked clean as-soon-as the beginning 💪 Cool to have reused the hooks and their corresponding queries.

* Good effort for trying to make the workspace image count being as efficient as possible 👍 

* I've enjoyed taking some time with you and @valerian to mob the migration!

Thanks to our efforts and the work you've done, we'll now be able to get the remaining Stripe stuff to be done with quick and scalable tasks 🚀 

If you're OK to merge the changes, you can squash the branch with a title such as:

```text
Support choosing workspace plan and track workspace workspace consumption and Stripe status
```

PR #918, opened 2022-02-24
Import / Export classes as CSV in a dataset

### LINKED ISSUE #905: Import / Export classes as CSV in a dataset

valerian:
## Work performed

- Add a CSV export feature to dataset label classes.
- Allow creating multiple classes at once from a list of names.
- Add a "createManyLabelClasses" mutation.
- Enforce check of duplicate LabelClass names within a dataset at creating time.

## Results

- We can export/download classes on the dataset label classes pages, in the CSV format.
- Class creation form the dataset classes page is now made using a list of multiple classes.

## Caveats

When inserting multiple new label classes the cache of the current dataset is entirely evicted.

## Resolved issues

Resolves #905

### DIFF
--- a/typescript/common-resolvers/src/label-class.ts
+++ b/typescript/common-resolvers/src/label-class.ts
@@ -2,6 +2,7 @@ import { v4 as uuidv4 } from "uuid";
 import type {
   LabelClass,
   MutationCreateLabelClassArgs,
+  MutationCreateManyLabelClassesArgs,
   MutationUpdateLabelClassArgs,
   MutationReorderLabelClassArgs,
   MutationDeleteLabelClassArgs,
@@ -75,14 +76,6 @@ const createLabelClass = async (
     user,
   });
 
-  // Since we don't have any constraint checks with Dexie
-  // we need to ensure that the datasetId matches some
-  // entity before being able to continue.
-  await throwIfResolvesToNil(
-    `The dataset id ${datasetId} doesn't exist.`,
-    repository.dataset.get
-  )({ id: datasetId }, user);
-
   const labelClassId = id ?? uuidv4();
   const now = new Date();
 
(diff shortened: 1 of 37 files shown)


typescript/web/src/components/dataset-classes/dataset-classes-export.ts
```diff
@@ -0,0 +1,20 @@
+import { LabelClassWithShortcut } from "./types";
+import {
+  getDatasetClassesExportName,
+  stringifyCsvAsync,
+} from "../../utils/export";
+
+export const exportDatasetClasses = async (
```
bbenoist, line 7: Can we have a unit-test for this function?

typescript/web/src/components/dataset-classes/dataset-classes-export.ts
```diff
@@ -0,0 +1,20 @@
...
+  datasetSlug: string,
+  classes: LabelClassWithShortcut[]
+) => {
```
bbenoist, line 10: Can we unit-test this code?

typescript/web/src/components/dataset-classes/insert-many-classes-modal/insert-many-classes-modal.tsx
```diff
@@ -0,0 +1,158 @@
+import { ApolloError } from "@apollo/client";
```
bbenoist, line 1: I've not seen any Story, Test or Cypress test for this new component. Can we have them?

typescript/web/src/utils/export/csv-stringify-async.ts
```diff
@@ -0,0 +1,18 @@
...
+): Promise<string> =>
+  new Promise((resolve, reject) => {
+    stringifyCsv(rows, options, (error, output) => {
+      if (error) {
+        reject(error);
+      } else {
+        resolve(output);
+      }
+    });
+  });
```
bbenoist, line 18: The same code already exists in `common-resolvers/src/export/export-to-csv.ts`. Can you please merge both?

bbenoist (changes requested):

typescript/db/src/repository/index.ts
```diff
@@ -177,13 +178,45 @@ export const repository: Repository = {
...
+        (await prisma.labelClass.findFirst({
+          where: {
+            OR: labelClasses.map((labelClass) => ({
+              name: labelClass.name,
+              datasetId: labelClass.datasetId,
+            })),
+          },
+        })) !== null;
+      if (nameExists)
+        throw new Error("A label class with the same name already exists");
```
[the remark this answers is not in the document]
valerian: Yes, and we have already talked about this, but we need to do it in a dedicated issue, and see what to do with the already existing duplicates

typescript/common-resolvers/src/label-class.ts
```diff
@@ -106,6 +99,48 @@ const createLabelClass = async (
...
+    user,
+  });
+  const nowIso = new Date().toISOString();
+  const newColors = getManyNextClassColors(
+    existingLabelClasses.map((item) => item.color),
+    newLabelClasses.length
+  );
+  const newLabelClassesEntities = newLabelClasses.map(
+    (newLabelClass, index) => ({
+      id: newLabelClass.id ?? uuidv4(),
```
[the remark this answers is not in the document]
bbenoist: But we don't need optimistic responses for this case. We should wait for the creation to complete and show an error in the modal. As for the workspace and dataset upserts.

typescript/db/src/repository/index.ts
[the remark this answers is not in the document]
bbenoist: > but we need to do it in a dedicated issue

Then we shouldn't address the issue at all in this PR.

> see what to do with the already existing duplicates

Remove them and replace references to their ID with the ID of the entity that's been kept.

typescript/common-resolvers/src/label-class.ts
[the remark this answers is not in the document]
valerian: My bad I already removed the uuid that used to be generated before the GQL query, this one is for another reason, and is done like that everywhere else in the common resolvers: it's to be able to return the re-query the DB using those IDs to return the inserted items data.
But yeah it's not the right way, at first I thought it was a "lazy" mistake from the person that coded our "addMany" repository function, since Prisma should return those IDs in the first place.

But no ... https://github.com/prisma/prisma/issues/8131
Also relevant ... https://www.reddit.com/r/node/comments/mip2lz/how_the_f_does_anyone_use_prisma_in_production/

typescript/web/src/components/dataset-classes/insert-many-classes-modal/insert-many-classes-modal.tsx
valerian: Story and Test done, but not Cypress: we currently don't do any Cypress tests on the dataset classes list tab, I think we should open a dedicated issue for that

PR #923, opened 2022-02-28
Add freehand drawing

### LINKED ISSUE #631: Do not store uploaded COCO zip to labelflow-images

### LINKED ISSUE #639: Make the freehand drawing tool more accessible via a dedicated mode
**Is your feature request related to a problem? Please describe.**

As of now, the freehand mode is somewhat hidden in the polygon tool (press shift when drawing).

**Describe the solution you'd like (import visuals when possible)**

Add a new **Freehand** tool.

Icon could be the BoxIcon `pencil`:
image

**Describe alternatives you've considered**

None

**Additional context**

Canny request: https://labelflow.canny.io/admin/feedback/feature-requests/p/add-free-hand-drawing-tool?boards=feature-requests

Jordanlelay:
## Work performed

- Add freehand drawing tool

## Resolved issues

Closes #639

### DIFF
--- a/typescript/web/src/components/labeling-tool/drawing-tool-bar/drawing-tool.tsx
+++ b/typescript/web/src/components/labeling-tool/drawing-tool-bar/drawing-tool.tsx
@@ -379,20 +379,6 @@ export const DrawingTool = () => {
                 </Box>
               </ToolSelectionPopoverItem>
               <ToolSelectionPopoverItem
-                name="Auto Polygon"
-                shortcut={keymap.toolIog.key}
-                selected={selectedTool === Tools.IOG}
-                onClick={() => {
-                  setSelectedTool(Tools.IOG);
-                  setIsPopoverOpened(false);
-                }}
-                ariaLabel="Auto polygon Tool"
-              >
-                <Box ml="2">
-                  <ChakraIoColorWandOutline size="1.3em" />
-                </Box>
-              </ToolSelectionPopoverItem>
-              <ToolSelectionPopoverItem
                 name="Freehand"
                 shortcut={keymap.toolFreehand.key}
                 selected={selectedTool === Tools.FREEHAND}
@@ -404,6 +390,20 @@ export const DrawingTool = () => {
               >
                 <Box ml="2">
                   <ChakraBiPencil size="1.3em" />
+                </Box>
+              </ToolSelectionPopoverItem>
+              <ToolSelectionPopoverItem
+                name="Auto Polygon"
+                shortcut={keymap.toolIog.key}
+                selected={selectedTool === Tools.IOG}
+                onClick={() => {
+                  setSelectedTool(Tools.IOG);
+                  setIsPopoverOpened(false);
+                }}
+                ariaLabel="Auto polygon Tool"
+              >
+                <Box ml="2">
+                  <ChakraIoColorWandOutline size="1.3em" />
                 </Box>
               </ToolSelectionPopoverItem>
             </Box>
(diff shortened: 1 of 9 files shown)


bbenoist (changes requested):
Although it contributes to increase the big chunks already there, code looks good at a first sight. Cool to the the feature finally there 🍾 

However, although it's quite meaningless to add a Storybook story for this tool, there remains the Jest unit-tests and Cypress e2e tests which are missing.
- [ ] Can you please add them?

Also, I've found a bug yesterday while I was testing the PR. It happened while switching between tools. The freehand tool was selected but I couldn't draw, it acted as-if the current tool was the selection-tool. But I couldn't find how to reproduce today.
**_Nothing blocker_** as-long-as it can't be reproduced but I had to warn you about it.

nicOwlas:
Cool! Tested it, works like a charm (except the alerts on the invalid geometries but that are already there for polygons)

Small request, not directly linked to this issue. Could you edit the text in the shortcut menu for the auto polygon? There are still references to IOG and Auto Annotate instead of Auto Polygon
image

Edited text:
- Switch between default selection and Auto Polygon edition tool
- Validate the edition of an Auto Polygon label

typescript/web/src/keymap.ts
```diff
@@ -23,6 +23,11 @@ export const keymap: Keymap = {
     description: "Polygon tool",
     category: "Tools",
   },
+  toolFreehand: {
+    key: "h",
+    description: "Freehand tool",
```
bbenoist, line 28: @nicOwlas is the `h` shorcut OK to you?
_(I'd use `f` but it's quite subjective)_

Jordanlelay: I would have too, but already taken by `Enter full screen` 😕

bbenoist: > F11 for windows

Yes but on most laptops, the F keys are now mapped some hardware features by default (e.g. volume, brightness, etc). Unless you change some settings in the BIOS, it's needed to also press the `Fn` key.

So `f` is often used as an alternative on some websites like YouTube...

typescript/web/src/components/labeling-tool/drawing-tool-bar/drawing-tool.tsx
```diff
@@ -371,6 +392,20 @@ export const DrawingTool = () => {
                   <ChakraIoColorWandOutline size="1.3em" />
                 </Box>
               </ToolSelectionPopoverItem>
+              <ToolSelectionPopoverItem
```
bbenoist, line 395: @nicOwlas The tool appears after Auto-polygon which I find somewhat counter-intuitive as it's a manual tool. Shouldn't we place it above it?

bbenoist: Yes

typescript/web/src/keymap.ts
nicOwlas: I did not know that `f` was also working on Youtube!

Lets keep `h` for freehand, it's still easy enough to remember.

bbenoist (approved):
It's all good to me now 👏

Very good idea to reuse EXAMPLE_POLYGON in the Cypress test! Thanks!

I've just pushed a simplification of the Cypress test I've made during the code-review. You can take a look at the commit and directly merge the PR if you're OK with its content :rocket:

nicOwlas:
I was about to say that we should put a feature flag to not release immediately but lets release it if ready, we will just communicate on it next week

Jordanlelay:
> It's all good to me now 👏
> 
> Very good idea to reuse EXAMPLE_POLYGON in the Cypress test! Thanks!
> 
> I've just pushed a simplification of the Cypress test I've made during the code-review. You can take a look at the commit and directly merge the PR if you're OK with its content 🚀

Yes it's cleaner like that 👍

PR #934, opened 2022-03-04
Add labels name and hotkey to hide them

### LINKED ISSUE #933: Display class name in the middle of the labels

Jordanlelay:
## Work performed

- Display labels name in the middle of the geometry
- Add a hotkey to hide labels name, geometry, classification tags

## Caveats

- Now that we can hide geometries from the labels, it's possible to add some without being able to see them
- Users should have a visual way to tell in which mode they currently are
- Today the hotkey is `x`, we should use something more user friendly. Discussions are in progress

## Resolved issues

Closes #933


nicOwlas:
Functional tests 👍 👏 
- Could you put the button above the full screen one?
- For the shortcut text I would write: "Toggle the visibility of labels' name and geometry"

bbenoist (approved):
Good work! I've made minor changes but everything looked functionnally OK.

Apart from factorization or renames to "better"-fit the architecture big-picture present and future, here is a brief summary of the changes I've made:
* I've prevented the state from being set twice when setting both view options to true
* Importing SVG files was not supported yet by Jest yet so I've added the needed configuration following svgr official recommendations.
* As you've seen on Slack, I've found a way to wire Chakra font family and size to the label text so that the gui feels _slightly_ more consistent.
* @niOwlas SVG files were not 24x24 as with the SVG icons we usually have. I've readjusted them to match this size and we had to tweak the Icon component but it doesn't feel perfect yet.

nicOwlas:
> * @niOwlas SVG files were not 24x24 as with the SVG icons we usually have. I've readjusted them to match this size and we had to tweak the Icon component but it doesn't feel perfect yet.

I thought the size was set on your side. No problem to ensure raw icons are 24x24. Let me know if you need me to re-size them on my side. As I've added the `A` in the bottom right, the standard size is changed.

typescript/web/src/components/labeling-tool/view-tool-bar/view-tool-bar.tsx
```diff
@@ -11,7 +11,7 @@ export const ViewToolbar = ({
 }) => {
   return (
     <>
-      <HideTool />
+      <ViewModeButton />
```
Jordanlelay, line 14: Should be named `ViewModeTool` as the rest of the components in this page.

typescript/web/src/connectors/labeling-state/index.ts
```diff
@@ -162,20 +152,25 @@ export const useLabelingStore = create<LabelingState>(
     showLabelsGeometry: true,
     showLabelsName: true,
     toggleViewMode: () => {
+      // Change labels visibility (geometry and name).
```
Jordanlelay, line 155: Putting the comment only here will avoid it to be discoverable in the editor when used.

bbenoist: Although I understand why you wanted to comment-out how it works, the thing is that:
* The implementation will evolve in a not-that-far future (we've already spoken about it during the last DSM) and if we put the comment on the type declaration, they are great chances that we'll forget to update it as we'll probably won't need to change the function typings (and won't read the comment it would had).
* It's not that important to know between which view mode it toggles at an upper-level in the code.
* If we want to have more details about what happens in one of our own functions, we're used to navigate to its implementation ; where I've moved the details you've provided 😉

typescript/web/src/components/labeling-tool/view-tool-bar/view-tool-bar.tsx
bbenoist: Actually the whole `*-tool-bar` directories and their content are not that well named IMO.
But it's no problem, I'll name it ViewModeTool for now as you suggested and we'll do a more-general cleanup later.

PR #940, opened 2022-03-07
In UpsertClassModal, don't show an error when new class name equals the current one

### LINKED ISSUE #866: Editing a class name in the class tab displays a wrong alert
In the class tab, when you edit the name of a class, it displays an alert saying
> This name is already taken

The name class conflicts should be checked on every classes of the dataset but the one being edited. What do you think?

## Tasks

- [ ] Try to merge with Workspace name input code

image

Jordanlelay:
## Work performed

- Fix incorrect error message when editing label class name

## Results

The modal used for editing label classes won't display an error when first opened. After changing the name of the class once, it has the same behavior as before.

## Resolved issues

Closes #866


typescript/web/src/components/dataset-classes/dataset-classes.context.tsx
```diff
@@ -28,6 +28,8 @@ export interface DatasetClassesState {
   loading: boolean;
   onReorder: OnReorderCallback;
   onExportClasses(): void;
+  setHasEditedClassName(value: boolean): void;
+  hasEditedClassName: boolean;
```
bbenoist, line 32: We don't need to have a state (and trigger extra renders) for this.
Instead, we should test whether the input value is equal to the `editClass?.name`.

typescript/web/src/components/dataset-classes/dataset-classes.tsx
```diff
@@ -12,12 +12,17 @@ import {
...
+  const {
+    loading,
+    labelClasses,
+    editClass,
+    setEditClass,
+    setHasEditedClassName,
+  } = useDatasetClasses();
+  const handleCloseEditModal = useCallback(() => {
+    setHasEditedClassName(false);
+    setEditClass(undefined);
```
bbenoist, line 24: As stated above, this will trigger 2 renders.

bbenoist (changes requested):

bbenoist (approved):
Thanks for the fix!
I've just pushed a very minor cleanup on top of your work 🧹
PR is good to be _squashed_ and merged once the CI goes green 🚀

PR #941, opened 2022-03-07
Add a border to the website hero video

bbenoist:
## Work performed

Border was missing on /website hero video

## Results

Website hero video now has a small gray border

## Problems encountered

No

## Caveats

No

## Resolved issues

No

## Newly raised issues

No

### DIFF
--- a/typescript/web/src/components/website/hero.tsx
+++ b/typescript/web/src/components/website/hero.tsx
@@ -7,7 +7,7 @@ import {
   Heading,
   Stack,
   Text,
-  useColorModeValue as mode,
+  useColorModeValue,
 } from "@chakra-ui/react";
 import { RiGithubFill } from "react-icons/ri";
 import { APP_GITHUB_URL } from "../../constants";
@@ -18,7 +18,7 @@ const CatchPhrase = () => (
   <Heading
     as="h1"
     size="3xl"
-    color={mode("gray.600", "gray.300")}
+    color={useColorModeValue("gray.600", "gray.300")}
     py="6"
     fontWeight="extrabold"
     letterSpacing="tight"
@@ -32,7 +32,7 @@ const CatchPhrase = () => (
 
 const Description = () => (
   <Text
-    color={mode("gray.600", "gray.400")}
+    color={useColorModeValue("gray.600", "gray.400")}
     mt="4"
     fontSize="lg"
     fontWeight="medium"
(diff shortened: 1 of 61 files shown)


bbenoist:
CC @geoffreyvancassel

typescript/web/src/components/website/hero.tsx
```diff
@@ -7,7 +7,7 @@ import {
   Heading,
   Stack,
   Text,
-  useColorModeValue as mode,
+  useColorModeValue,
```
Jordanlelay, line 10: You should let it as it is for now to stay consistent with the majority of our code, or refactor every other place that make that rename

Jordanlelay (changes requested):

typescript/web/src/components/website/hero.tsx
bbenoist: On such invasive and "soft" relations things (_the `mode` name is local to the file importing the function and `useColorModeValue` is imported in a lot of files_), I try to be more progressive and do the changes over the time.

For this, specific case, it's not always renamed as `mode` in the existing code. For example, I've nearly always used `useColorModeValue` since my arrival and replaced `mode` as I saw it in the code I had to change.

This because:
* React hooks function names must begin with `use`. We're bypassing some verifications by not using the prefix.
* `mode` is a way too generic word IMO. If the function name length was actually a problem (_e.g. makes the code less readable_), then I'd use a more precise name such as `useColorMode` or eventually `useColor`.
* We can still use a proxy function in `utils` with a shorter name instead of forcing an implicit consistency (_using the `mode` name in each file importing `useColorModeValue`_) such as the one you've found.

But you're right that it's inconsistent for now👍 I'll do the switch to `useColorModeValue` once for all in an additional commit on this PR so that we can forget about `mode` 🚀

bbenoist: Done

Jordanlelay (approved):

nicOwlas:
I tested in dark mode. Everything looks good.

I catched an error on the auto polygon modal: the link to the documentation is wrong and leads to a non existing page in GitBook.
- Current link: https://docs.labelflow.ai//labelflow/labelling-interface/label-types/auto-polygons (note the `//`)
- Should be: https://docs.labelflow.ai/labelflow/labelling-interface/label-types/auto-polygons

image

PR #944, opened 2022-03-08
Update image design in dataset's image gallery

### LINKED ISSUE #853: Update image card in image gallery
## Tasks

- [x]  update image card in gallery (fill not fit)
- [x] Default state
  - [x] Put a black gradient covering the lower third of the image with 40% transparency
  - [x] Name is semi-bold fond
- [x] Hover state
  - [x] 40% transparency layer above the entire card
  - [x] Add the bin in the top right
- [x] on hover the bin button darkens

## Mockup
- [Link to Figma](https://www.figma.com/file/4J9xyfF5v4B98MuKVO0Y8V/LabelFlow-mockups?node-id=4907%3A19401)
image

Jordanlelay:
## Work performed

- Update image design in dataset's image gallery

## Resolved issues

Closes #853


nicOwlas:
> When the card not hovered, the end of the gradient shadow feels too brutal IMO.

Agree. Suggested fix: use black-alpha 700 (black 64% opacity) at the bottom gray-300 opacity 0% as the top color instead of black 0%.

>  When the card is hovered, the transparency does not matches the one at the bottom of the not hovered card.

Ok to change for black-alpha 700 for the hover card. It was initially done on purpose, as the card was very dark if using black-alpha 700 on hover but I tested on other images and it looks fine to me.

> I'd add a very quick transition on opacity when mouse is entering/leaving the card.

Ok. If it's "immediate"

> I'd harmonize the dataset card in this PR too (at least use the gradient shadow) as I think that it looks very weird to have both versions on the same release version.

In another issue. I did a quick mockup, I'll create the issue.

> The card size ratio feels not wide-enough to me.

Agree, can we make it wider? The mockup shows a 1.5 ratio (DSLR and high end image aspect ratio, smartphone picture is 1.33), on the preview at max I see a 1.4.

> The remove icon feels too small.

Use 16x16 as in the mockup

> The remove button is opaque when hovered. Maybe that it's nicer to keep it a little bit transparent and only have it 100% opaque when pressed 🤷‍♂️

Agree with transparency it would be nicer. The mockup shows 20% opacity.
 
 
image

Jordanlelay:
@nicOwlas I've fixed the reviews except for the aspect ratio one which is not as immediate as one could think. I'm working on it now.

Ahead of a *potential* comment saying "it should be as easy as using chakra's `<AspectRatio />`", no it's not. As it's actually a css trick using a padding, it breaks everything below it and especially the overlay I've coded. I'm looking for another way to increase the ratio

typescript/web/src/components/dataset-images-list/image-card.tsx
```diff
@@ -58,65 +43,110 @@ const DeleteButton = () => {
...
+const OverlayTopRow = () => (
+  <HStack
+    visibility="hidden"
+    justifyContent="flex-end"
+    w="100%"
+    sx={{
+      ".imageCard:hover &": {
+        visibility: "visible",
+      },
+    }}
```
bbenoist, line 84: You can use the Chakra `role="group"` and `_groupHover` for this specific case.
https://github.com/chakra-ui/chakra-ui/discussions/2386#discussioncomment-118607

If it still doesn't work with it, we should bind the correct events of ImageCard to the `setOn` or `setTrue` return properties of the `useBoolean` hook and forward the boolean value to the component.

typescript/web/src/components/dataset-images-list/image-card.tsx
```diff
@@ -58,65 +43,110 @@ const DeleteButton = () => {
...
   );
 };
 
+const ImageCardContent = () => (
+  <Box
+    className="imageCard"
+    position="relative"
+    cursor="pointer"
+    sx={{
+      ".imageCard:hover .imageOverlay, .imageOverlay:hover": {
```
bbenoist, line 138: Can't we really have the same with no CSS?

bbenoist (changes requested):

bbenoist (approved):
Code was good and the card seemed nice.

I've made a few cleanup, added a story for ImageCard and made sure some edge-cases were supported (no thumbnails, error loading the image, etc.). You can look at the commit description for details about the changes.

If you're OK with the content, then you can directly _squash_ and merge this PR on main 🚀

PR #955, opened 2022-03-14
Make sure that the workspace switcher is always visible in breadcrumbs

### LINKED ISSUE #943: Add min width to workspaceSwitcher

Jordanlelay:
## Work performed

- Fix breadcrumbs width to make workspace switcher always visible

## Resolved issues

Closes #943


bbenoist (changes requested):
I have a problem when going to the workspaces page:
image

bbenoist (approved):
Everything seems to be working well now on both desktop and mobile. Good job!

bbenoist (changes requested):
Sorry for the late denial but, while validating the Chromatic snapshots before merging, I've found some invalid ones. Mostly due to an extra left margin added to some of the elements.
You can look at the Chromatic review for details.

Jordanlelay:
@bbenoist I don't understand what is "invalid" in the chromatic screenshots, is it the half character displacement ?

bbenoist:
@Jordanlelay Yes, that should be the only issue IIRC:

image

Jordanlelay:
> @Jordanlelay Yes, that should be the only issue IIRC

@bbenoist the stories do not match what is displayed in front if you look at it

Front:
Capture d’écran 2022-03-18 à 17 02 47

Story:
Capture d’écran 2022-03-18 à 17 05 41

Jordanlelay:
> If there is a problem with the stories themselves, then the stories have to be fixed 😉

IMO there is nothing to fix as the only case where the component is used does not behave as in the stories. So it does not make sense to have stories that do not test real component behavior right ?

To say it differently, I can fix the said stories while introducing problems in the app as they don't act the same way

PR #957, opened 2022-03-15
Bulk delete of images

### LINKED ISSUE #854: Bulk delete of images
## Tasks

- [x] Add a checkbox icon on the image card
- [x] When selected the card is in hover mode
- [x] Display the bulk action buttons at the top
- [x] Drop shadow below the bulk action buttons (2 pixels, blur 4, gray-200 as Figma)
- [x] Please respect button font colors, size and height of tool bar (let me know if you need to change things compared to the mockup)
- [x] `Select` button states
  - [x] If 0 images are selected the `Select` button displays `Select all`.
  - [x] If one or more images are selected the `Select` button displays `Deselect all`
  - [x] `Select all` selects the item displayed on the page only
  - [x] We store the list of image ids selected. We display the delete action on every pages if at least one item is in the list of selected items
  - [x] Clear the list of selected items when the user quits the gallery
- [x] Display confirmation modal when user presses delete (same as today but with the updated text).
  - [x] Delete the selected images together with their labels
- [x] Make the top nav bar always visible
- [x] Keep the navigation bar always visible
- [x] Display the number of selected items in the bottom right (cf mockup)

**Extra**
- [x] Reduce height of top navbar (where we have the breadcrumbs) from 72 to 56 pix (I don't know why and when it was increased to 72). Mockup shows 56. Reduce padding to `--chakra-space-2`

## Mockup

- [Bulk action buttons](https://www.figma.com/file/4J9xyfF5v4B98MuKVO0Y8V/LabelFlow-mockups?node-id=4989%3A20244)
image

- [Image gallery with bulk delete buttons](https://www.figma.com/file/4J9xyfF5v4B98MuKVO0Y8V/LabelFlow-mockups?node-id=4907%3A19401)
image

- [Confirmation modal Figma](https://www.figma.com/file/4J9xyfF5v4B98MuKVO0Y8V/LabelFlow-mockups?node-id=4607%3A17195)
image

Depends on #853

Jordanlelay:
## Work performed

- Add bulk selection of images
- Update delete image modal
- Reduce top bar height
- Reduce dataset tab height
- Move items number on right of footer toolbar
- Display how many images are selected in bottom left corner

## Resolved issues

Closes #854


typescript/web/src/components/dataset-images-list/delete-many-images-modal.tsx
```diff
@@ -0,0 +1,68 @@
...
+  const [deletedImagesIds, { loading }] = useMutation<
+    DeleteManyImagesMutation,
+    DeleteManyImagesMutationVariables
+  >(DELETE_MANY_IMAGES_MUTATION, {
+    update: (cache) => cache.evict({ id: `Dataset:${datasetId}` }),
+  });
+
+  const handleDeleteButtonClick = async () => {
+    await flushPaginatedImagesCache();
+    await deletedImagesIds({
```
valerian, line 45: You should add a `awaitRefetchQueries: true` here, so that the modal only closes when the updated list of images is loaded in the gallery. Same for the 1 image delete modal.

valerian (dismissed):

nicOwlas:
I tested with 1000+ images, it's a bit slow but usable.

One problem though: when I select one image, the 1000 images transition to "hover" mode, is it on purpose ? If so why ? I would keep the default card on every images not selected.

PR #965, opened 2022-03-24
Use TextLink in website footer LinkGrid and cookies banner

Jordanlelay:
## Work performed

Fix incorrect nesting of chakra's Link in NextLink


bbenoist (approved):
@Jordanlelay Thanks for helping on this issue too! I've pushed a commit which reuses the TextLink you've previously added so that the code gets _a little-bit_ smaller. Let me know if there is a problem with it and we could still revert/change it :rocket:

PR #977, opened 2022-04-01
Set useSuccessToast default duration to 5s

bbenoist:
## Work performed

I've reduced the default successful toast duration to 5s

## Results

Toast should take less time to disappear

## Problems encountered

No

## Caveats

No

## Resolved issues

No

## Newly raised issues

No

### DIFF
--- a/typescript/web/src/components/toast.ts
+++ b/typescript/web/src/components/toast.ts
@@ -8,13 +8,19 @@ import { isNil } from "lodash/fp";
 import { useCallback } from "react";
 import { getApolloErrorMessage } from "../utils/get-apollo-error-message";
 
+const DEFAULT_TOAST_DURATION = 5000;
+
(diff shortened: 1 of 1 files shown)


typescript/web/src/components/toast.ts
```diff
@@ -37,7 +39,7 @@ const getToastOptions = (
 export const useToast = (...hookArgs: ToastFnParameters) => {
   const toastOptions = getToastOptions(hookArgs);
   const toast = useChakraToast({
-    // Make description actually optional since when title is **not** empty:
+    // Make description actually optional since, when title is **not** empty:
```
Jordanlelay, line 42:
```suggestion
    // Make description actually optional because if title is **not** empty:
```

Jordanlelay (changes requested):

PR #978, opened 2022-04-04
Store AiAssistant ID in db when creating new labels

bbenoist:
## Work performed

I've added the AI Assistant ID in the `Label` entity of the DB schema.

## Results

We can now acknowledge that a label has been created with an AiAssistant.

## Caveats

DB relations are not complete as we don't store AI Assistant in it yet.

## Resolved issues

_No issue opened for it AFAIK but it's related with #958_

## Newly raised issues

No


Jordanlelay (approved):

Jordanlelay (changes requested):
tests are failing

bbenoist:
@Jordanlelay It should be OK now

